PR #1, opened 2020-06-23
Setting up directories and adding required player files.

juliareichel:
-Set up the directory tree and copied over HTML and javascript player files


SIMID/examples/maps/player/simid_player.html
```diff
@@ -0,0 +1,117 @@
+ <html>
```
damargulis, line 1: You shouldn't need any of the player/ files.

You can continue to use the player hosted from the simid repo, and only host your creatives here

damargulis, line 1: Add some text to the maps/README file. Can be short, just explain what the creative will be and what files will go in that dir

juliareichel: done!

juliareichel: deleted them

juliareichel:
> This might be a good start, but for the appengine to work correctly, the folders, tests and html/js documents should be organized very differently than

maps/simid_protocol.js
```diff
@@ -0,0 +1,388 @@
+/**
+ * Contains logic for sending mesages between the SIMID creative and the player.
+ * Note: Some browsers do not support promises and a more complete implementation
+ *       should consider using a polyfill.
```
damargulis, line 4: Add a note giving the url where this file originally came from.

maps/README.txt
```diff
@@ -0,0 +1,6 @@
+The goal of this project is to create a sample interactive ad creative for SIMID. 
+Our aim is to design a creative using the Google Maps API that provides users a 
+map of nearby locations where they could find the advertised product.
+
+The files in this directory include:
```
damargulis, line 5: You don't need to actually list the files, just a description of in general what will go here.
I.e. something like `Contains files for SIMID creative that will populate the iframe`

maps/README.txt
```diff
@@ -0,0 +1,6 @@
+The goal of this project is to create a sample interactive ad creative for SIMID.
```
damargulis, line 1: Looks like you accidentally added .README.txt.swp, be sure to remove

You can add the line `*.swp` to your .gitignore file and that should prevent accidentally adding swp files

juliareichel: Deleted.

maps/README.txt
juliareichel: Got it! Fixed

maps/simid_protocol.js
juliareichel: Added

PR #7, opened 2020-06-24
Set up basic files

juliareichel:
- Includes Basic SIMID Creative JS file
- HTML Maps Creative Template
- JS skeleton for Maps Creative


maps/creatives/simid_maps.html
```diff
@@ -0,0 +1,9 @@
+<html>
+  <head>
+    <script src="../simid_protocol.js"> </script>
```
damargulis, line 3: I think you can move the simid_protocol class into this folder.
It was a folder above in the SIMID repo because its being used by the player code as well, but since there is no player here, its only being used by the creatives and can go into that folder

juliareichel: Is this true even though we are using the same player?

damargulis: Yeah cause you won't be copying over any of the player code into this repo.  You can use the player being hosted by the simid repo, but no need to keep this file in a shared location, since it isn't being shared in this repo but is only being used by creatives

juliareichel: okay great fixed

damargulis:
Sorry, just noticed this -- it should probably go creatives/maps not maps/creatives

If you end up building another creative, then it would go in creatives/otherCreativeType

damargulis:
Then, protocol and base_creative can go in creatives/ with SimidMapCreative going in creative/maps

PR #12, opened 2020-06-25
Set up Jest for Javascript Unit Testing

damargulis:
This converts the current code to use es6 modules, which are required to use the Jest framework. This also adds a base_simid_creative.test.js file with a few sample unit tests, showing how to check expectations, and a simple example of using mocks.


package.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage": "https://github.com/googleinterns/step33-2020#readme",
+  "devDependencies": {
+    "jest": "^26.1.0"
+  },
+  "dependencies": {
+    "babel-jest": "^26.1.0",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2"
+  },
+  "babel": {
+    "plugins": ["transform-es2015-modules-commonjs"]
```
Kaleab470, line 27: Where do we use babel in this code?

damargulis: We don't directly, but jest is using it under the hood.  Jest isn't set up to directly understand es6 modules, but this babel plugin converts the es6 style modules into commonjs style modules, which jest is then able to understand.

PR #13, opened 2020-06-26
Servlet files

### LINKED ISSUE #10: Add Initial Servlet Files

Kaleab470:
Closes #10

### DIFF
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -14,10 +14,4 @@ public class FindLocationServlet extends HttpServlet {
     response.setContentType("text/html;");
     response.getWriter().println("<h1>Find Location</h1>");
   }
-
-  @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-
-    response.sendRedirect("/find-location");
-  }
 }
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -10,15 +10,16 @@ import javax.servlet.http.HttpServletResponse;
 public class FindLocationServlet extends HttpServlet {
 
  /**
-  * Given the correlator, this will update the database to reflect that the user clicked 'fiind nearby location'.
+  * Given the correlator, this will update the database to reflect that the user
+  * clicked 'find nearby location'.
   *
-  * @param  request  This includes a correlator for the current user.
-  * @param  response An empty 200 response 
+  * request - This includes a correlator of type String for the current user.
+  * response - An empty 200 response 
   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - change default response
+    //TODO - implement the rest of this function (@Kaleab)
     response.setStatus(HttpServletResponse.SC_OK); 
   }
 }
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -1,4 +1,3 @@
-
 package com.google.sps.servlets;
 
 import java.io.IOException;
@@ -7,13 +6,13 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-@WebServlet("/initialize")
-public class InitializeServlet extends HttpServlet {
+@WebServlet("/findLocation")
+public class FindLocationServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("text/html;");
-    response.getWriter().println("<h1>Initialize</h1>");
+    response.getWriter().println("<h1>Find Location</h1>");
   }
 
   @Override
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -9,6 +9,12 @@ import javax.servlet.http.HttpServletResponse;
 @WebServlet("/find-location")
 public class FindLocationServlet extends HttpServlet {
 
+ /**
+  * Given the correlator, this will update the database to reflect that the user clicked 'fiind nearby location'.
+  *
+  * @param  request  This includes a correlator for the current user.
+  * @param  response An empty 200 response 
+  */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("text/html;");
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/sps/servlets/InitializeServlet.java
```diff
@@ -0,0 +1,24 @@
+
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/initialize")
```
juliareichel, line 10: what is this servlet going to be handling?

Kaleab470: It is where the initial entry is made on the server side. Sent on init of creative

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,23 @@
...
+@WebServlet("/find-location")
+public class FindLocationServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+    response.getWriter().println("<h1>Find Location</h1>");
+  }
+
+  @Override
```
damargulis, line 18: Why include these doPost functions? If the intention is to only use get requests, then only define get requests

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,23 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/find-location")
+public class FindLocationServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+    response.getWriter().println("<h1>Find Location</h1>");
```
damargulis, line 15: This is just temporary right?
Leave a TODO to fix this then.


Or better yet, set it up to return a 200 response instead of some dummy html (still include the todo to implement)

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,23 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/find-location")
```
damargulis, line 9: include descriptions for each of these .. when will they be requested? What params will they need to accept?

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: so if i understand correctly, we are making get requests to these endpoints with the correlator as a parameter. is this a good way/style to send stuff to the server without using post bc a get request is not meant for that purpose?

damargulis: IMO I think its ok to use get requests for these even though you aren't really "getting" any information, since its pretty standard to use get requests for tracking pings in VAST (often servers will respond with a 1x1 pixel image, or just an empty 200 response).  If you use the simid spec's ReportTracking function, it will trigger a get request.  

I think its fine to just use get requests for this then, and you can just remove the doPost definitions

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,24 @@
...
+public class FindLocationServlet extends HttpServlet {
+
+ /**
+  * Given the correlator, this will update the database to reflect that the user clicked 'fiind nearby location'.
+  *
+  * @param  request  This includes a correlator for the current user.
```
damargulis, line 15: These param tags are fine to leave, but aren't actually usually needed for java unless you feel like you need to explain the variable more.

Theyre more useful in javascript, where you cant know the type of the inputs just by reading the code.

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,24 @@
...
+  * @param  response An empty 200 response 
+  */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    //TODO - change default response
```
damargulis, line 21: I think this will probably stay as a 200 response (unless you had something else you wanted to return?), but the TODO should rather say to actually implement the rest of this function (fetch and update the database?)
Add tag either a person or github issue in it

src/main/java/com/google/sps/servlets/GrantLocationServlet.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/grant-location")
+public class GrantLocationServlet extends HttpServlet {
+
+ /**
+  * Given the correlator, this will update the database to reflect that the user allowed their location to be accessed.
```
damargulis, line 13: (hard to tell on github you might be ok already)
Be sure to limit lines to <=100 characters (80 for javascript but 100 for java)

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/find-location")
+public class FindLocationServlet extends HttpServlet {
+
+ /**
+  * Given the correlator, this will update the database to reflect that the user clicked 'fiind nearby location'.
```
damargulis, line 13: Either in this description, or in the param comment if you choose to leave it in, you should actually specify the format the correlator should come it. So whats the query param its attached to? Is it a string? Number? boolean?

src/main/java/com/google/sps/servlets/GrantLocationServlet.java
Kaleab470: i just set up an extension for that so it should be taken care of for the future!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: I removed them

PR #17, opened 2020-06-26
Servlet implementation

### LINKED ISSUE #16: Get backend code working

Kaleab470:
Closes #16

### DIFF
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -18,7 +18,7 @@ public class DBUtilities {
   * clicked on the corresponding metric.
   *
   * @param correlator A correlator for the current user.
-  * @param propertyToUpdate The property that is to be updated
+  * @param propertyToUpdate The property that is to be updated.
   */
   public static void setToTrue(String correlator, String propertyToUpdate) {
     
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -10,7 +10,7 @@ import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
-import com.google.sps.servlets.Propery;
+import com.google.sps.servlets.Property;
 
 public class DBUtilities {
 
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -14,7 +14,7 @@ import com.google.sps.servlets.Property;
 
 public class DBUtilities {
 
-  public final String INTERACTION_TABLE = "Interactions";
+  public static final String INTERACTION_TABLE = "Interactions";
 
  /**
   * Given an HTTP request object, this method will get the correlator parameter
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -40,14 +40,17 @@ public class DBUtilities {
   public static void setToTrue(String correlator, String propertyToUpdate) {
     
     final Filter correlatorFilter =  new FilterPredicate(Property.CORRELATOR, FilterOperator.EQUAL, correlator);
-    final Query impressionQuery = new Query(INTERACTION_TABLE).setFilter(correlatorFilter);
+    final Query interactionQuery = new Query(INTERACTION_TABLE).setFilter(correlatorFilter);
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
+    PreparedQuery filteredImpression = datastore.prepare(interactionQuery);
 
-    for (Entity impression : filteredImpression.asIterable()) {
-      impression.setProperty(propertyToUpdate, true);
-      datastore.put(impression);  // override the existing entity
+    // throws a tooManyResultsException if more than one entry exists with same correlator
+    Entity currentInteraction = filteredImpression.asSingleEntity();
+
+    if (currentInteraction != null) {
+      currentInteraction.setProperty(propertyToUpdate, true);
+      datastore.put(currentInteraction);  // override the existing entity
     }
   }
 }
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -8,27 +8,10 @@ import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.datastore.Query.FilterOperator;
-import javax.servlet.http.HttpServlet;
-import javax.servlet.http.HttpServletRequest;
-import com.google.sps.servlets.Property;
 
 public class DBUtilities {
 
   public static final String INTERACTION_TABLE = "Interactions";
-
- /**
-  * Given an HTTP request object, this method will get the correlator parameter
-  * and return it. Returns an empty string if correlator is not given.
-  */
-  public static String getCorrelator(HttpServletRequest request) {
-    String correlator = request.getParameter(Property.CORRELATOR);
-
-    if (correlator == null) {
-      return "";
-    }
-
-    return correlator;
-  }
  
  /**
   * Given the correlator and the property to update, this will update the database to reflect that the user
--- a/src/main/java/com/google/sps/servlets/DBUtilities.java
+++ b/src/main/java/com/google/sps/servlets/DBUtilities.java
@@ -17,8 +17,8 @@ public class DBUtilities {
   * Given the correlator and the property to update, this will update the database to reflect that the user
   * clicked on the corresponding metric.
   *
-  * correlator - A correlator of type String for the current user.
-  * propertyToUpdate - A property of type String that is to be updated
+  * @param correlator A correlator for the current user.
+  * @param propertyToUpdate The property that is to be updated
   */
   public static void setToTrue(String correlator, String propertyToUpdate) {
     
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -21,7 +21,7 @@ public class FindLocationServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    String correlator = request.getParameter("correlator");
+    final String correlator = request.getParameter("correlator");
 
     if (correlator == null) {
       response.setStatus(HttpServletResponse.SC_NOT_FOUND);
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -16,8 +16,8 @@ public class FindLocationServlet extends HttpServlet {
   * Given the correlator, this will update the database to reflect that the user
   * clicked 'find nearby location'.
   *
-  * request - This includes a correlator of type String for the current user.
-  * response - An empty 200 response 
+  * @param request This includes a correlator for the current user.
+  * @param response A 200 status or 400 error
   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -5,7 +5,7 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.sps.servlets.Shared;
+import com.google.sps.servlets.DBUtilities;
 
 
 @WebServlet("/find-location")
@@ -21,14 +21,14 @@ public class FindLocationServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    final String correlator = Shared.getCorrelator(request);
+    final String correlator = DBUtilities.getCorrelator(request);
 
     if (correlator.isEmpty()) {
       response.setStatus(HttpServletResponse.SC_NOT_FOUND);
       return;
     }
 
-    Shared.updateDatabase(correlator, Shared.FIND_NEAREST_LOCATION);
+    DBUtilities.setToTrue(correlator, Property.FIND_NEAREST_LOCATION);
 
     response.setStatus(HttpServletResponse.SC_OK); 
   }
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -6,6 +6,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.servlets.DBUtilities;
+import com.google.sps.servlets.Propery;
 
 
 @WebServlet("/find-location")
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -29,22 +29,11 @@ public class FindLocationServlet extends HttpServlet {
     
     String correlator = request.getParameter("correlator");
 
-    //TODO: find a more graceful way to handle this
     if (correlator == null){
-      throw new IOException();
+      response.setStatus(HttpServletResponse.SC_NOT_FOUND);
     }
-    
-    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
-    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
 
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
-
-    //only runs once
-    for (Entity impression : filteredImpression.asIterable()) {
-      impression.setProperty("clicksFindNearestLocation", true);
-      datastore.put(impression);  //override the existing entity
-    }
+    Constants.updateDatabase(correlator, "clicksFindNearestLocation");
 
     response.setStatus(HttpServletResponse.SC_OK); 
   }
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -5,8 +5,12 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.datastore.Query.FilterOperator;
 
@@ -31,7 +35,7 @@ public class FindLocationServlet extends HttpServlet {
     }
     
     Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
-    Query impressionQuery = new Query("Impressions").setFilter(keyFilter);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -42,7 +42,7 @@ public class FindLocationServlet extends HttpServlet {
 
     //only runs once
     for (Entity impression : filteredImpression.asIterable()) {
-      impressions.setProperty("clicksFindNearestLocation", true);
+      impression.setProperty("clicksFindNearestLocation", true);
     }
 
     //TODO - implement the rest of this function (@Kaleab)
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -5,7 +5,7 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.sps.servlets.Constants;
+import com.google.sps.servlets.Shared;
 
 
 @WebServlet("/find-location")
@@ -21,9 +21,9 @@ public class FindLocationServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    final String correlator = request.getParameter("correlator");
+    final String correlator = Shared.getCorrelator(request);
 
-    if (correlator == null) {
+    if (correlator.isEmpty()) {
       response.setStatus(HttpServletResponse.SC_NOT_FOUND);
       return;
     }
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -6,7 +6,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.servlets.DBUtilities;
-import com.google.sps.servlets.Propery;
+import com.google.sps.servlets.Property;
 
 
 @WebServlet("/find-location")
--- a/src/main/java/com/google/sps/servlets/FindLocationServlet.java
+++ b/src/main/java/com/google/sps/servlets/FindLocationServlet.java
@@ -28,7 +28,7 @@ public class FindLocationServlet extends HttpServlet {
       return;
     }
 
-    Constants.updateDatabase(correlator, "clicksFindNearestLocation");
+    Shared.updateDatabase(correlator, "clicksFindNearestLocation");
 
     response.setStatus(HttpServletResponse.SC_OK); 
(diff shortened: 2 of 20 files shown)


src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - implement the rest of this function (@Kaleab)
+    String correlator = request.getParameter("correlator");
+
+    //TODO: find a more graceful way to handle this
```
damargulis, line 32: tag someone / a github issue

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - implement the rest of this function (@Kaleab)
+    String correlator = request.getParameter("correlator");
```
damargulis, line 30: Make const, for any being reused in multiple files, you can create a constants file that exports all the consts you will need

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
+    }
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
```
damargulis, line 38: "Impressions" - make const

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
+
+    //only runs once
```
damargulis, line 43: What does this mean?

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+    for (Entity impression : filteredImpression.asIterable()) {
+      impression.setProperty("clicksFindNearestLocation", true);
+      datastore.put(impression);  //override the existing entity
```
damargulis, line 46: nit. comments should have a space after the `//`

src/main/java/com/google/sps/servlets/InitializeServlet.java
```diff
@@ -20,7 +23,24 @@
...
-    //TODO - implement the rest of this function (@Kaleab)
+    String correlator = request.getParameter("correlator");
+
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
+    }
+
+    Entity impressions = new Entity("Impressions");
+    impressions.setProperty("correlator", correlator);
```
damargulis, line 34: all these should be consts

src/main/java/com/google/sps/servlets/MapInteractServlet.java
```diff
@@ -19,7 +27,25 @@
...
+    }
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
+
+    //only runs once
+    for (Entity impression : filteredImpression.asIterable()) {
```
damargulis, line 44: This whole thing is repeated in every file, see if you can extract it into a shared helper

damargulis:
Also, add tests for all these endpoints

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - implement the rest of this function (@Kaleab)
+    String correlator = request.getParameter("correlator");
+
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
```
ryanthompson591: you should add a string message to the IOException.

Or by contrast, you could just throw no exception at all.  It looks like what you are trying to do is validate incoming data.

If the incoming data is not in the correct format it could be just spam, or not from your service.  In that case just exiting this function and doing nothing seems like the correct course of action.

Perhaps make another function called validRequest.

if (!validRequest(request)) {
  // The request is not from our service or has invalid data.
  return;
}

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
+
+    //only runs once
+    for (Entity impression : filteredImpression.asIterable()) {
+      impression.setProperty("clicksFindNearestLocation", true);
```
ryanthompson591: use a const for clicksFindNearestLocation.  Can you also comment why you are adding this field to your data?

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
+    }
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
```
ryanthompson591: In the advertising world impression means something very specific (a user has viewed an ad).  Correct me if I'm wrong but is this service only called on a click? If so use the word click instead of impression here and elsewhere.

src/main/java/com/google/sps/servlets/GrantLocationServlet.java
```diff
@@ -5,6 +5,14 @@
...
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.FilterOperator;
 
 @WebServlet("/grant-location")
 public class GrantLocationServlet extends HttpServlet {
```
ryanthompson591, line 11: Validate that we get this data at all.  I'm curious if granting location access is something that is between google maps and the user or not.

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -5,6 +5,14 @@
...
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.FilterOperator;
 
 @WebServlet("/find-location")
 public class FindLocationServlet extends HttpServlet {
```
ryanthompson591, line 12: comment all these classes.

something like.

/**
 * This servlet is called by the ad whenever the user clicked on the find location button on the ad.
 */

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +27,25 @@
...
+
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
+    }
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
ryanthompson591: Feel free to break this into a separate function here and elsewhere.  Took me a while to figure this all out.  A much more readable function could look like this:

public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
  if (!validRequest(request)) {
   // If something other than the creative calls this service return invalid.
   response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
   return;
  }
  storeData(request);
  response.setStatus(HttpServletResponse.SC_OK); 
}

/**
  * Modifies the DB for this ad to show that the user had clicked on "find nearest location"
  * on the ad.
  */
private storeData(HttpServletRequest request) {
  ...
}

src/main/java/com/google/sps/servlets/SkipToContentServlet.java
```diff
@@ -19,7 +27,25 @@
...
+    //TODO: find a more graceful way to handle this
+    if (correlator == null){
+      throw new IOException();
+    }
+    
+    Filter correlatorFilter =  new FilterPredicate("correlator", FilterOperator.EQUAL, correlator);
+    Query impressionQuery = new Query("Impressions").setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
```
ryanthompson591: There's a lot of repeated code here. It might make you wish to make an abstract class.  That's one way to go, but IMO a utility class might be best.

src/main/java/com/google/sps/servlets/InitializeServlet.java
Kaleab470: When you say these, are you referring to the properties themselves?

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: I return a 404 response code now!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done!

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Done! I structured it this way now

src/main/java/com/google/sps/servlets/MapInteractServlet.java
Kaleab470: Done! Made a shared class

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: Added it to the class javadoc

src/main/java/com/google/sps/servlets/FindLocationServlet.java
Kaleab470: done!

src/main/java/com/google/sps/servlets/Shared.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+
+public class Shared {
```
ryanthompson591, line 14: Shared is a very uncommon name for a utilities class.  This is fine if you want to keep for now, but a more common way to do this would be to just have two files.

Put all the constants into a single file.
Put all the utility functions into another file.

Call the class something like Key, Constant or Property.

that way you would call it like:
Property.CORRELATOR

src/main/java/com/google/sps/servlets/GrantLocationServlet.java
```diff
@@ -18,8 +19,16 @@
...
-    //TODO - implement the rest of this function (@Kaleab)
+ 
+    final String correlator = Shared.getCorrelator(request);
+
+    if (correlator.isEmpty()) {
+      response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+      return;
+    }
+
+    Constants.updateDatabase(correlator, Shared.GRANTS_LOCATION);
```
ryanthompson591, line 30: I like this change. I wonder if your files are out of date.  Did you call this Shared?

src/main/java/com/google/sps/servlets/Shared.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+ 
+ /**
+  * Given the correlator and the property to update, this will update the database to reflect that the user
+  * clicked on the corresponding metric.
+  *
+  * correlator - A correlator of type String for the current user.
+  * propertyToUpdate - A property of type String that is to be updated
+  */
+  public static void updateDatabase(String correlator, String propertyToUpdate) {
```
ryanthompson591, line 45: A more common name might be DatabaseUtilities, DBUtil, Utilities, DataHandler.

A more common way to name methods is to try and be more specific about what it actually does for example setToTrue.

That way your function call would be
DataBaseUtil.setToTrue(correlator, Constant.FIND_NEAREST_LOCATION);

Pretty clear exactly what that line does, whereas updateDatabase is ambiguous (update to what)

src/main/java/com/google/sps/servlets/GrantLocationServlet.java
Kaleab470: Oops good catch!

src/main/java/com/google/sps/servlets/Shared.java
Kaleab470: Okay great. I have named one class DBUtilities and the other Property to make it clear

src/main/java/com/google/sps/servlets/Shared.java
Kaleab470: done!

src/main/java/com/google/sps/servlets/DBUtilities.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class DBUtilities {
+
+  public static final String INTERACTION_TABLE = "Interactions";
+
+ /**
+  * Given an HTTP request object, this method will get the correlator parameter
+  * and return it. Returns an empty string if correlator is not given.
+  */
+  public static String getCorrelator(HttpServletRequest request) {
```
damargulis, line 23: This doesn't seem like it belongs in a class called DBUtils, since it has nothing to do with the database

src/main/java/com/google/sps/servlets/DBUtilities.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return correlator;
+  }
+ 
+ /**
+  * Given the correlator and the property to update, this will update the database to reflect that the user
+  * clicked on the corresponding metric.
+  *
+  * correlator - A correlator of type String for the current user.
```
damargulis, line 37: Use javaDoc style if you are going to reference the parameters
```
/**
 * description...
 * @param correlator Description of the correlator parameter.
 */
```

You also don't need to include the type ("string") in java, since its evident from reading the code.  Thats much more helpful in js, where you can't tell from the code what the type would be

src/main/java/com/google/sps/servlets/DBUtilities.java
```diff
@@ -0,0 +1,53 @@
...
+  */
+  public static void setToTrue(String correlator, String propertyToUpdate) {
+    
+    final Filter correlatorFilter =  new FilterPredicate(Property.CORRELATOR, FilterOperator.EQUAL, correlator);
+    final Query impressionQuery = new Query(INTERACTION_TABLE).setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(impressionQuery);
+
+    for (Entity impression : filteredImpression.asIterable()) {
```
damargulis, line 48: What if this has either 0 or >1 results? Maybe that should be some kind of error?

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -19,7 +22,15 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - implement the rest of this function (@Kaleab)
+    final String correlator = DBUtilities.getCorrelator(request);
+
+    if (correlator.isEmpty()) {
+      response.setStatus(HttpServletResponse.SC_NOT_FOUND);
```
damargulis, line 28: This would execute if they don't pass in a correlator right?
But you're returning a 404, which is usally used for when an entity isn't found, meaning like that url doesn't exist.  This should probably be a 400 - Bad Request.

A 404 could be used if they send in a correlator, but it can't be found in the db anywhere

src/main/java/com/google/sps/servlets/DBUtilities.java
Kaleab470: Where would you suggest I put this? I wasn't a fan of putting it in a separate class with just one method but I also want it to be shared.

damargulis: Putting it in its own class is fine (RequestUtils?), that way if you end up adding more functions you already have a place to put them.

Alternatively, you could rename this class to just utils, but imo itd be better to keep them separate

src/main/java/com/google/sps/servlets/DBUtilities.java
Kaleab470: The asSingleEntity() throws a tooManyResultsException if more than one entry exists with same correlator. I have changed my code to use that method now!

src/main/java/com/google/sps/servlets/DBUtilities.java
Kaleab470: Okay, done!

src/main/java/com/google/sps/servlets/DBUtilities.java
damargulis: Doesn't look like this was done

Kaleab470:
> lgtm!
> Just make sure to adjust the comment on that last file

The requestUtils file?

damargulis:
The dbUtils class

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    final Filter correlatorFilter =  new FilterPredicate(Property.CORRELATOR, FilterOperator.EQUAL, "Person1");
+    final Query interactionQuery = new Query(INTERACTION_TABLE).setFilter(correlatorFilter);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery filteredImpression = datastore.prepare(interactionQuery);
+
+    Entity currentInteraction = filteredImpression.asSingleEntity(); 
+
+    assertEquals(originalInteraction.toString(), currentInteraction.toString());
```
juliareichel, line 96: What exactly will this toString() show? Is it comparing "true" to "true" or "Find_Nearest_Location" to "Find_Nearest_Location"?

src/test/java/com/google/sps/servlets/FindLocationServletTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    when(request.getParameter("correlator")).thenReturn("Person1");
+
+    new FindLocationServlet().doGet(request, response);
+
+    assertEquals(HttpServletResponse.SC_OK, response.getStatus());
+  }
+
+  @Test
+  public void checkIfBadRequestStatusSentNull() {
```
juliareichel, line 26: above you started the name with test and here you start with check, choose one and stay consistent

src/test/java/com/google/sps/servlets/FindLocationServletTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    when(request.getParameter("correlator")).thenReturn(null);
+
+    new FindLocationServlet().doGet(request, response);
+
+    assertEquals(HttpServletResponse.SC_BAD_REQUEST, response.getStatus());
+  }
+
+  @Test
+  public void checkIfBadRequestStatusSentEmpty() {
```
juliareichel, line 38: ditto

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
```diff
@@ -0,0 +1,98 @@
...
+    DBUtilities.setToTrue("Person1", Property.FIND_NEAREST_LOCATION);
+
+    assertEquals(1, dataStore.prepare(new Query(DBUtilities.INTERACTION_TABLE)).countEntities(withLimit(10)));
+  }
+
+  @Test
+  public void checkIfPropertyUpdatedCorrectly() {
+    DatastoreService datastore = setDatabaseUp();
+    
+    DBUtilities.setToTrue("Person1", Property.GRANTS_LOCATION);
```
juliareichel, line 61: maybe specify that GRANTS_LOCATION is only an example of a property that you can test with this particular unit test, but that it's applicable for ALL property tests

src/test/java/com/google/sps/servlets/RequestUtilsTest.java
```diff
@@ -0,0 +1,40 @@
...
+  public void testCorrelatorInvalid() {
+    HttpServletRequest request = mock(HttpServletRequest.class);    
+
+    when(request.getParameter("correlator")).thenReturn("");
+    
+    assertEquals(RequestUtils.getCorrelator(request), "");
+  }
+
+  @Test
+  public void testCorrelatorValid() {
```
juliareichel, line 33: this test has the same name as the first (perhaps make them different)

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
Kaleab470: okay thanks

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
Kaleab470: its comparing the entire entities to each other to make sure nothing was changed

src/test/java/com/google/sps/servlets/FindLocationServletTest.java
Kaleab470: good catch!

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
juliareichel: got it, so all of the field are compared

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
```diff
@@ -0,0 +1,99 @@
...
+    PreparedQuery filteredImpression = newDatastore.prepare(interactionQuery);
+
+    Entity currentInteraction = filteredImpression.asSingleEntity(); 
+
+    Assert.assertEquals(true, (boolean) currentInteraction.getProperty(Property.GRANTS_LOCATION));
+  }
+
+  @Test
+  public void testNothingChangesWhenPersonNotFound() {
+    DatastoreService datastore = setDatabaseUp();
```
damargulis, line 84: You can call this in the setup function, instead of in each test

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
```diff
@@ -0,0 +1,99 @@
...
+    originalInteraction.setProperty(Property.SKIP_TO_CONTENT, false);
+    originalInteraction.setProperty(Property.RETURN_TO_AD, false);
+
+    dataStore.put(originalInteraction);
+
+    return dataStore;
+  }
+
+  @Test
+  public void testIfOnlyOneEntryMade() {
```
damargulis, line 55: I don't think I see what this is testing.  This would pass even if you never called anything DBUtils

Kaleab470: the idea for this test is to make sure the setToTrue function overrides the entry and doesn't end up making a duplicate entry with an updated field.

damargulis: Ahh gotcha makes sense.  Maybe just change the name of the test a bit to specify that.

`testSetToTrueDoesntCreateNewEntry` or something else a bit more specific.

For test names especially, since you never actually call them anywhere, theyre fine if theyre super long

src/test/java/com/google/sps/servlets/DBUtilitiesTest.java
Kaleab470: that's a good idea!

src/main/java/com/google/sps/servlets/DBUtilities.java
```diff
@@ -0,0 +1,39 @@
...
+public class DBUtilities {
+
+  public static final String INTERACTION_TABLE = "Interactions";
+ 
+ /**
+  * Given the correlator and the property to update, this will update the database to reflect that the user
+  * clicked on the corresponding metric.
+  *
+  * @param correlator A correlator for the current user.
+  * @param propertyToUpdate The property that is to be updated
```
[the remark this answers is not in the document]
Kaleab470: done.

src/main/java/com/google/sps/servlets/InitializeServlet.java
```diff
@@ -14,13 +20,40 @@
...
+
+    // this doesn't handle an entry iwth the same correlator
+    // the initialize route would send the key to the entity it creates
+    // the js file would use that for the rest of the session
+
+    // uuid
+
+    // maybe ask ryan?
+    // 3: hash function on client side
+    // 5: don't handle collisions and assume they don't happen often and error out
```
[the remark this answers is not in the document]
Kaleab470: done.

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -13,13 +17,21 @@
...
+  * @param request This includes a correlator for the current user.
+  * @param response A 200 status or 400 error.
   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     
-    //TODO - implement the rest of this function (@Kaleab)
+    final String correlator = RequestUtils.getCorrelator(request);
+
+    if (correlator.isEmpty()) {
```
kristen-m, line 28: I noticed that you're making this correlator check in most of the Java files, I don't know how feasible it would be but is there any way you could make this into one helper method that you can access multiple times? It could help with readability in the long term

Kaleab470: I did think about putting it in the utilities file but i think the way its structured now( if not valid send 404, else set to true and send 200) makes more sense logically to me when reading it. could be wrong tho

PR #20, opened 2020-06-30
Handling errors

juliareichel:
-Moved ad parameter checking to onInit instead of onStart
-Dealt with various incorrect/nonexistent input for ad parameters

### DIFF
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -22,24 +22,18 @@ export default class SimidMapCreative extends BaseSimidCreative {
 
   /** @override */
   onInit(eventData) {
+    this.handleAdParams_(eventData);
+    super.onInit(eventData);
+  }
+
+  /**
+   * Checks validity of ad parameters and rejects with proper message if invalid.
+   * @param eventData an object that contains information details for a particular event
+   *   such as event type, unique Ids, creativeData and environmentData.
+   * @private 
+  */ 
+  handleAdParams_(eventData) {
     this.creativeData = eventData.args.creativeData;
-    this.environmentData = eventData.args.environmentData;
-
-
-    this.videoState = {
-      currentSrc:'',
-      currentTime: -1, // Time not yet known
-      duration: -1, // duration unknown
-      ended: false,
-      muted: this.environmentData.muted,
-      paused: false,
-      volume: this.environmentData.volume,
-      fullscreen: false,
-    }
-
-    /**
-     * Error handeling for empty or incorrect ad paramters.
-     */
     let adParams = "";
     if (this.creativeData.adParameters == "") {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
@@ -49,7 +43,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
     try {
       adParams = JSON.parse(this.creativeData.adParameters);
     } catch (exception) {
-      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.MESSAGES_NOT_FOLLOWING_SPEC, 
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.CREATIVE_INTERNAL_ERROR, 
         message: "Invalid JSON input"});
     }
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
@@ -59,9 +53,6 @@ export default class SimidMapCreative extends BaseSimidCreative {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
         message: "Required field 'SearchQuery' not found"});
     }
-
-    const marker = adParams[AdParamKeys.MARKER];
-    this.simidProtocol.resolve(eventData, {});
   }
 
   /** @override */
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -1,5 +1,6 @@
 
 import BaseSimidCreative from '../base_simid_creative.js';
+import {CreativeErrorCode} from '../constants.js';
 
 const AdParamKeys = {
   BUTTON_LABEL: 'buttonLabel',
@@ -20,27 +21,53 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /** @override */
-  onStart(eventData){
-    super.onStart(eventData);
+  onInit(eventData) {
+    this.creativeData = eventData.args.creativeData;
+    this.environmentData = eventData.args.environmentData;
+
+
+    this.videoState = {
+      currentSrc:'',
+      currentTime: -1, // Time not yet known
+      duration: -1, // duration unknown
+      ended: false,
+      muted: this.environmentData.muted,
+      paused: false,
+      volume: this.environmentData.volume,
+      fullscreen: false,
+    }
+
+    /**
+     * Error handeling for empty or incorrect ad paramters.
+     */
+    let adParams = "";
+    if (this.creativeData.adParameters == "") {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Ad parameters not found"});
+    }
+
     try {
-      const adParams = JSON.parse(this.creativeData.adParameters);
-    } catch (e) {
-      this.sendErrorMessage_(CreativeErrorCode.MESSAGES_NOT_FOLLOWING_SPEC);
+      adParams = JSON.parse(this.creativeData.adParameters);
+    } catch (exception) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.MESSAGES_NOT_FOLLOWING_SPEC, 
+        message: "Invalid JSON input"});
     }
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
-    //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
-    const marker = adParams[AdParamKeys.MARKER];
 
     if (searchQuery == undefined) {
-      this.sendErrorMessage_(CreativeErrorCode.UNSPECIFIED);
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Required field 'SearchQuery' not found"});
     }
 
-    this.specifyButtonFeatures_(buttonLabel);
+    const marker = adParams[AdParamKeys.MARKER];
+    this.simidProtocol.resolve(eventData, {});
   }
 
-  sendErrorMessage_(message) {
-    this.simidProtocol.sendMessage(message);
+  /** @override */
+  onStart(eventData, buttonLabel) {
+    super.onStart(eventData);
+    this.specifyButtonFeatures_(buttonLabel);
   }
 
   /**
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -22,15 +22,27 @@ export default class SimidMapCreative extends BaseSimidCreative {
   /** @override */
   onStart(eventData){
     super.onStart(eventData);
-    //ToDo(juliareichel@): handle invalid JSON and param errors
-    const adParams = JSON.parse(this.creativeData.adParameters);
+    try {
+      const adParams = JSON.parse(this.creativeData.adParameters);
+    } catch (e) {
+      this.sendErrorMessage_(CreativeErrorCode.MESSAGES_NOT_FOLLOWING_SPEC);
+    }
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
     //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
     const marker = adParams[AdParamKeys.MARKER];
+
+    if (searchQuery == undefined) {
+      this.sendErrorMessage_(CreativeErrorCode.UNSPECIFIED);
+    }
+
     this.specifyButtonFeatures_(buttonLabel);
   }
- 
+
+  sendErrorMessage_(message) {
+    this.simidProtocol.sendMessage(message);
+  }
+
   /**
    * Sets the text of the Find Nearest button and assigns it a click functionality.
    * @param {string=} buttonLabel Refers to the value given to the BUTTON_LABEL key in 
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -34,12 +34,13 @@ export default class SimidMapCreative extends BaseSimidCreative {
   */ 
   validateAndParseAdParams_(eventData) {
     this.creativeData = eventData.args.creativeData;
-    let adParams = "";
+
     if (this.creativeData.adParameters == "") {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
         message: "Ad parameters not found"});
     }
 
+    let adParams = "";
     try {
       adParams = JSON.parse(this.creativeData.adParameters);
     } catch (exception) {
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -22,8 +22,8 @@ export default class SimidMapCreative extends BaseSimidCreative {
 
   /** @override */
   onInit(eventData) {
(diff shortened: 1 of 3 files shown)


damargulis:
Looks like this broke some tests, can you fix first

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+      muted: this.environmentData.muted,
+      paused: false,
+      volume: this.environmentData.volume,
+      fullscreen: false,
+    }
+
+    /**
+     * Error handeling for empty or incorrect ad paramters.
+     */
+    let adParams = "";
```
ryanthompson591, line 43: This function has gotten a little complex.
How about breaking out all validation code into a separate function?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    if (this.creativeData.adParameters == "") {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Ad parameters not found"});
+    }
+
+    try {
+      adParams = JSON.parse(this.creativeData.adParameters);
+    } catch (exception) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.MESSAGES_NOT_FOLLOWING_SPEC,
```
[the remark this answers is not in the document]
juliareichel: makes sense-changed to internal creative error

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+      ended: false,
+      muted: this.environmentData.muted,
+      paused: false,
+      volume: this.environmentData.volume,
+      fullscreen: false,
+    }
+
+    /**
+     * Error handeling for empty or incorrect ad paramters.
+     */
```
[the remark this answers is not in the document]
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
-    //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
+
+    if (searchQuery == undefined) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Required field 'SearchQuery' not found"});
+    }
+
     const marker = adParams[AdParamKeys.MARKER];
```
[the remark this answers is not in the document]
juliareichel: I am not using it in this function- but I plan to use it in a future function

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /** @override */
-  onStart(eventData){
-    super.onStart(eventData);
-    //ToDo(juliareichel@): handle invalid JSON and param errors
-    const adParams = JSON.parse(this.creativeData.adParameters);
+  onInit(eventData) {
+    this.creativeData = eventData.args.creativeData;
```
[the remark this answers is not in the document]
juliareichel: They need to be defined because I use them in my tests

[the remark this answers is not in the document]
juliareichel: I use creativeData.adParameters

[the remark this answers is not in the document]
juliareichel: I guess I can delete environmentData for now

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: done

src/main/webapp/creatives/maps/simid_map_creative.js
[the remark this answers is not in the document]
juliareichel: I can get rid of it here then

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    this.environmentData = eventData.args.environmentData;
+
+
+    this.videoState = {
```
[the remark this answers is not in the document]
juliareichel: okay- I will call on super.onInit(eventData) to build the object

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+        message: "Invalid JSON input"});
+    }
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
-    //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
-    const marker = adParams[AdParamKeys.MARKER];
+
+    if (searchQuery == undefined) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Required field 'SearchQuery' not found"});
```
Kaleab470, line 54: Consider making this searchQuery (camelcased) as opposed to SearchQuery bc the parameter is case sensitive and the error message might be misleading as to how to name it.

Or better yet just use AdParamKeys.SEARCH_QUERY here as a string template so that if we do decide to change the parameter name, we don't have to change a whole lot and just the declaration at the top of the file.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  /** @override */
+  onStart(eventData, buttonLabel) {
+    super.onStart(eventData);
     this.specifyButtonFeatures_(buttonLabel);
   }
- 
+
   /**
    * Sets the text of the Find Nearest button and assigns it a click functionality.
    * @param {string=} buttonLabel Refers to the value given to the BUTTON_LABEL key in
```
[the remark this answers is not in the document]
juliareichel: signifies that the parameter is a default param

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: switched to AdParamKeys.SEARCH_QUERY

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    super.onInit(eventData);
+  }
+
+  /**
+   * Checks validity of ad parameters and rejects with proper message if invalid.
+   * @param eventData an object that contains information details for a particular event
+   *   such as event type, unique Ids, creativeData and environmentData.
+   * @private 
+  */ 
+  handleAdParams_(eventData) {
```
ryanthompson591, line 35: you could even name this function more specifically

validateAdParams_
or
validateAndParseAdParams_

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    if (this.creativeData.adParameters == "") {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: "Ad parameters not found"});
+    }
+
+    try {
+      adParams = JSON.parse(this.creativeData.adParameters);
+    } catch (exception) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.CREATIVE_INTERNAL_ERROR, 
+        message: "Invalid JSON input"});
```
damargulis, line 47: nit.
Invalid JSON input for AdParameters

Just to be a little more specific

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    }
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
-    //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
-    const marker = adParams[AdParamKeys.MARKER];
+
+    if (searchQuery == undefined) {
```
damargulis, line 52: If searchQuery is an empty string `""`, do you want to consider that an error?
If so, you can use `if (!searchQuery)`, and that will handle all falsey values (null, undefined, empty string, etc.)

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: changed

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -24,13 +24,14 @@ test('testing button text updates from ad params', () => {
...
         messageId: 1,
         sessionId: "test-session-id",
         timestamp: 0,
         type: "SIMID:Player:startCreative",  
     }
-    testMap.onStart(startData);
+    testMap.onStart(startData, buttonLabel);
 
     const button = document.getElementById("findNearest");
     expect(button.innerText).toBe('Find Nearest Place');
```
damargulis, line 37: Add some tests with invalid data, and make sure that the message is rejected properly

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,46 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /** @override */
-  onStart(eventData){
-    super.onStart(eventData);
-    //ToDo(juliareichel@): handle invalid JSON and param errors
-    const adParams = JSON.parse(this.creativeData.adParameters);
+  onInit(eventData) {
+    this.validateAndParseAdParams_(eventData);
+    super.onInit(eventData);
```
damargulis, line 26: If there is invalid data, then you'd send a reject message, but then this will still run and you'd send a resolve message anyway, right?

I think that would be a problem.  You probably need to find a way to only send resolve if all these functions pass properly. It might require some refactoring of base_simid_creative or simid_protocol (it might not, there is probably solutions that don't touch those as well)

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: good idea

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: done

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,49 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+
+  /**
+   * Checks validity of ad parameters and rejects with proper message if invalid.
+   * @param eventData an object that contains information details for a particular event
+   *   such as event type, unique Ids, creativeData and environmentData.
+   * @private 
+  */ 
+  validateAndParseAdParams_(eventData) {
+    this.creativeData = eventData.args.creativeData;
+    let adParams = "";
```
Kaleab470, line 37: Minor nit- you should declare this right above the try statement to help readability

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -24,13 +27,14 @@ test('testing button text updates from ad params', () => {
     };
     testMap.onInit(eventData);
 
+    const buttonLabel = "Place";
     const startData = {
         messageId: 1,
         sessionId: "test-session-id",
         timestamp: 0,
         type: "SIMID:Player:startCreative",  
     }
```
Kaleab470, line 36: This startData object seems to be repeated in another test so maybe consider declaring it a top level const for reuse

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: okay, moved

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: Since the eventData object is different in every test and therefore I declare it in the individual tests, I felt that having this startData object being declared as a top level const seems inconsistent. Since I only use this object twice, I didn't feel that it was a storage issue, but instead easier to read. Let me know your thoughts!

Kaleab470: Okay, sounds good!

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,50 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /** @override */
-  onStart(eventData){
-    super.onStart(eventData);
-    //ToDo(juliareichel@): handle invalid JSON and param errors
-    const adParams = JSON.parse(this.creativeData.adParameters);
+  onInit(eventData) {
+    this.validateAndParseAdParams_(eventData);
+    super.updateInternalOnInit(eventData);
```
damargulis, line 26: Since SimidMapCreative does not define a updateInternalOnInit function, you don't need super here, you can use `this.updateInternal..`

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,50 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  }
+
+  /**
+   * Checks validity of ad parameters and rejects with proper message if invalid.
+   * @param eventData an object that contains information details for a particular event
+   *   such as event type, unique Ids, creativeData and environmentData.
+   * @private 
+  */ 
+  validateAndParseAdParams_(eventData) {
+    this.creativeData = eventData.args.creativeData;
```
damargulis, line 36: If you swap the validate and updateInternal calls in onInit, you can remove this line

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,50 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
     const buttonLabel = adParams[AdParamKeys.BUTTON_LABEL]; 
-    //ToDo(juliareichel@): handle case where searchQuery is undefined
     const searchQuery = adParams[AdParamKeys.SEARCH_QUERY];
-    const marker = adParams[AdParamKeys.MARKER];
+
+    if (!searchQuery) {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
+        message: `Required field ${AdParamKeys.SEARCH_QUERY} not found`});
+    }
+    else {
```
damargulis, line 57: Its a good idea to reduce nesting whenever possible. Here you can do that by returning after each reject, and then you won't need the else clause here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -20,17 +21,50 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+      super.onInit(eventData);
```
damargulis, line 58: super.onInit will call updateInternalOnInit, so I don't think you also need it here.  Alternatively, leave the call here, but just call resolve here instead of super.onINit

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: good to know!

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: got it, switched to just calling resolve

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: swapped and removed

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: got rid of the else clause

PR #27, opened 2020-07-06
Moved button so not blocking play sign.

juliareichel:


kristen-m:
Do you know if we need to take into account different screen sizes for the creative? If we do that might be something to take into consideration for the button placement

juliareichel:
> Do you know if we need to take into account different screen sizes for the creative? If we do that might be something to take into consideration for the button placement

I am not sure, this is something we can discuss but for testing purposes I just wanted to move it out of the way

PR #29, opened 2020-07-06
Added buttons and simid requests.

juliareichel:
-created Skip To Content and Return To Ad buttons
-incorporated SIMID calls: REQUEST.PLAY, REQUEST.PAUSE, REQUEST.STOP

### DIFF
--- a/src/main/webapp/creatives/maps/simid_map.html
+++ b/src/main/webapp/creatives/maps/simid_map.html
@@ -3,7 +3,7 @@
     <link rel="stylesheet" type="text/css" href="simid_map.css">
   </head>
   <body>
-    <div id="container">
+    <div id="iframeContainer">
       <button id="findNearest"></button>
     </div>
     <script type="module">
--- a/src/main/webapp/creatives/maps/simid_map.html
+++ b/src/main/webapp/creatives/maps/simid_map.html
@@ -3,7 +3,7 @@
     <link rel="stylesheet" type="text/css" href="simid_map.css">
   </head>
   <body>
-    <div id="iframeContainer">
+    <div id="adContainer">
       <button id="findNearest"></button>
     </div>
     <script type="module">
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -37,8 +37,8 @@ export default class SimidMapCreative extends BaseSimidCreative {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
         message: "Ad parameters not found"});
         return;
-    }Fsk
-
+    }
+    
     let adParams = "";
     try {
       adParams = JSON.parse(this.creativeData.adParameters);
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -31,7 +31,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @param eventData an object that contains information details for a particular event
    *   such as event type, unique Ids, creativeData and environmentData.
    * @private 
-  */ 
+   */ 
   validateAndParseAdParams_(eventData) {
     if (this.creativeData.adParameters == "") {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
@@ -72,7 +72,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
    *   category and can be specified by the advertisers. If the value is not specified, 
    *   then BUTTON_LABEL's value will default to Location.
    * @private 
-  */   
+   */   
   specifyButtonFeatures_(buttonLabel = DEFAULT_BUTTON_LABEL) {
     const findNearestButton = document.getElementById('findNearest');
     findNearestButton.innerText = FIND_NEAREST_TEMPLATE_TEXT + buttonLabel;
@@ -82,7 +82,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
   /**
    * Requests player to pause, if accepted generate map.
    * @private 
-  */
+   */
   prepareCreative_() {
     //ToDo(kristenmason@): implement the Google Maps request access functionality
     findNearest.classList.add("hidden");
@@ -100,7 +100,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * Creates the Skip To Content and Return To Ad buttons once the user
    *   grants permission to access their location and the map appears.
    * @private 
-  */
+   */
   createMapState_() {
     const returnToAdButton = document.createElement("button");
     returnToAdButton.textContent = "Return To Ad";
@@ -124,7 +124,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @param {!Element} returnToAdButton Refers to the button that takes
    *   a user back to the video ad. 
    * @private 
-  */
+   */
   playAd_(returnToAdButton) {
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PLAY);
     returnToAdButton.classList.add("hidden");
@@ -134,7 +134,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
   /**
    * Returns to video content if user clicks on Skip To Content button.
    * @private 
-  */
+   */
   playContent_() {
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_SKIP);
   }
@@ -145,7 +145,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * TODO(kristenmason@): implement grant location access and modify
    * function to pass in current position (currently coords default to GooglePlex)
    * @private 
-  */
+   */
   loadMap_(coordinates = new google.maps.LatLng(37.422004,-122.081402)) { 
     const map = new google.maps.Map(document.getElementById('map'), {
       zoom: DEFAULT_ZOOM,
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -37,7 +37,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.UNSPECIFIED, 
         message: "Ad parameters not found"});
         return;
-    }
+    }Fsk
 
     let adParams = "";
     try {
@@ -84,9 +84,11 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   prepareCreative_() {
+    debugger;
     //ToDo(kristenmason@): implement the Google Maps request access functionality
     findNearest.classList.add("hidden");
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      debugger;
       this.createMapState_();
     }).catch(() => {
         const pauseErrorMessage = {
@@ -109,7 +111,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
 
     const skipAdButton = document.createElement("button");
     skipAdButton.textContent = "Skip Ad";
-    skipAdButton.id = "skipToContent";
+    skipAdButton.id = "skipAd";
     skipAdButton.onclick = () => this.playContent_();
 
     const adContainer = document.getElementById('adContainer');
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -84,11 +84,9 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   prepareCreative_() {
-    debugger;
     //ToDo(kristenmason@): implement the Google Maps request access functionality
     findNearest.classList.add("hidden");
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
-      debugger;
       this.createMapState_();
     }).catch(() => {
         const pauseErrorMessage = {
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -84,10 +84,8 @@ export default class SimidMapCreative extends BaseSimidCreative {
   */
   grantLocationAccess_() {
     //ToDo(kristenmason@): implement the Google Maps request access functionality
-    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
-      findNearest.classList.add("hidden");
-    }).catch(() => {
-      findNearest.classList.add("hidden");
+    findNearest.classList.add("hidden");
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).catch(() => {
       const pauseErrorMessage = {
         message: "WARNING: Request to pause ad failed",
       };
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -1,7 +1,6 @@
 
 import BaseSimidCreative from '../base_simid_creative.js';
-import {CreativeMessage} from '../constants.js';
-import {CreativeErrorCode} from '../constants.js';
+import {CreativeMessage, CreativeErrorCode} from '../constants.js';
 
 const AdParamKeys = {
   BUTTON_LABEL: 'buttonLabel',
@@ -53,14 +52,13 @@ export default class SimidMapCreative extends BaseSimidCreative {
   */
   grantLocationAccess_() {
     //ToDo(kristenmason@): implement the Google Maps request access functionality
-    try{
-      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
-        findNearest.classList.add("hidden");})
-    }
-    catch(exception){
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      findNearest.classList.add("hidden");
+    }).catch(() => {
       this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.PAUSE_NOT_HONORED, 
-        message: "Request to pause ad not honored"});
-    }
+      message: "Request to pause ad not honored"
+      });
+    });
     this.createButtonsMapState_();
   }
 
@@ -71,12 +69,12 @@ export default class SimidMapCreative extends BaseSimidCreative {
   */
   createButtonsMapState_() {
     const returnToAdButton = document.createElement("button");
-    returnToAdButton.innerHTML = "Return To Ad";
+    returnToAdButton.textContent = "Return To Ad";
     returnToAdButton.id = "returnToAd";
     returnToAdButton.onclick = () => this.playAd_(returnToAdButton); 
 
     const skipAdButton = document.createElement("button");
-    skipAdButton.innerHTML = "Skip Ad";
+    skipAdButton.textContent = "Skip Ad";
     skipAdButton.id = "skipToContent";
     skipAdButton.onclick = () => this.playContent_();
 
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -76,17 +76,19 @@ export default class SimidMapCreative extends BaseSimidCreative {
   specifyButtonFeatures_(buttonLabel = DEFAULT_BUTTON_LABEL) {
     const findNearestButton = document.getElementById('findNearest');
     findNearestButton.innerText = FIND_NEAREST_TEMPLATE_TEXT + buttonLabel;
-    findNearest.onclick = () => this.grantLocationAccess_();
+    findNearest.onclick = () => this.prepareCreative_();
   }
  
   /**
-   * Prompts the users to grant or deny access to their current location.
+   * Requests player to pause, if accepted generate map.
    * @private 
   */
-  grantLocationAccess_() {
+  prepareCreative_() {
+    debugger;
     //ToDo(kristenmason@): implement the Google Maps request access functionality
     findNearest.classList.add("hidden");
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      debugger;
       this.createMapState_();
     }).catch(() => {
         const pauseErrorMessage = {
@@ -121,6 +123,8 @@ export default class SimidMapCreative extends BaseSimidCreative {
 
   /**
    * Continues to play the ad if user clicks on Return To Ad button.
+   * @param {!Element} returnToAdButton Refers to the button that takes
+   *   a user back to the video ad. 
    * @private 
   */
   playAd_(returnToAdButton) {
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -80,9 +80,9 @@ export default class SimidMapCreative extends BaseSimidCreative {
     skipAdButton.id = "skipToContent";
     skipAdButton.onclick = () => this.playContent_();
 
-    const iframeContainer = document.getElementById('iframeContainer');
-    iframeContainer.appendChild(returnToAdButton);
-    iframeContainer.appendChild(skipAdButton);
+    const adContainer = document.getElementById('adContainer');
+    adContainer.appendChild(returnToAdButton);
+    adContainer.appendChild(skipAdButton);
   }
 
   /**
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -60,6 +60,11 @@ export default class SimidMapCreative extends BaseSimidCreative {
     })
   }
 
+  /**
+   * Creates the Skip To Content and Return To Ad buttons once the user
+   *   grants permission to access their location and the map appears.
+   * @private 
+  */
   createButtons_() {
     const returnToAdButton = document.createElement("button");
     returnToAdButton.innerHTML = "Return To Ad";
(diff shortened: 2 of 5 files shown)


src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -42,6 +43,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
   specifyButtonFeatures_(buttonLabel = DEFAULT_BUTTON_LABEL) {
     const findNearestButton = document.getElementById('findNearest');
     findNearestButton.innerText = FIND_NEAREST_TEMPLATE_TEXT + buttonLabel;
+
```
ryanthompson591, line 46: I just wouldn't add a newline unless you think it needs it.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,37 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      findNearest.classList.add("hidden");
+      //ToDo(kristenmason@): display map
+      this.createButtons_();
+    })
```
ryanthompson591, line 60: Consider adding a catch section here for if the player does not allow you to pause.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,37 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  }
+
+  createButtons_() {
```
ryanthompson591, line 68: createButtons is kind of a vague name.  I think this is for specifically buttons when the map is being accessed. If that's the case, either comment this function or better yet rename the function.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,37 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    const returnToAdButton = document.createElement("button");
+    returnToAdButton.innerHTML = "Return To Ad";
+    returnToAdButton.id = "returnToAd";
+    returnToAdButton.onclick = () => this.playAd_(returnToAdButton);
```
ryanthompson591, line 72: having a returnToAd button is super uncommon.  The more common thing to do is to just have a skip button only.

I wonder if what this is, is more of a "close map" button, perhaps it should be called that.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,37 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  }
+
+  playAd_(returnToAdButton) {
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PLAY);
+    returnToAdButton.classList.add("hidden");
+    //ToDo(kristenmason@): hide map
+  }
+
+  playContent_() {
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_STOP);
```
ryanthompson591, line 99: I suggest using request_skip instead of request_stop since we are skipping the ad right?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,37 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  }
+
+  createButtons_() {
+    const returnToAdButton = document.createElement("button");
+    returnToAdButton.innerHTML = "Return To Ad";
+    returnToAdButton.id = "returnToAd";
+    returnToAdButton.onclick = () => this.playAd_(returnToAdButton); 
+
+    const skipToContentButton = document.createElement("button");
+    skipToContentButton.innerHTML = "Skip To Content";
```
ryanthompson591, line 75: Most skip buttons just say "Skip" and a user understands what that is, perhaps you could say "Skip Ad" if you wanted.

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: Must have been extra from when I deleted the console.log. Got rid of it

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: I added jsdoc explaining the function in the second commit (maybe you didn't see it). I can change the name also

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: The reason we designed it this way is if users wish to return to the ad to watch the rest of the video and learn more about the product. IMO "Return to Ad" is more informative than just close map.

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: I can change that

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: okay -- I can change to Skip Ad

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: good idea!

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,50 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    skipToContentButton.id = "skipToContent";
+    skipToContentButton.onclick = () => this.playContent_();
+
+    const container = document.getElementById('container');
```
kristen-m, line 79: Do you think we could give this a more descriptive name than container? In another PR Kaleab suggested that in the html we call the div button_container for added clarity, I think it would be a good idea here as well

juliareichel: the div I am referring to here is literally the entire iframe container so I can call it that if you'd prefer?

src/main/webapp/creatives/maps/simid_map.html
```diff
@@ -3,7 +3,7 @@
     <link rel="stylesheet" type="text/css" href="simid_map.css">
   </head>
   <body>
-    <div id="container">
+    <div id="iframeContainer">
```
kristen-m, line 6: This is a much better name!! I definitely did not know that was originally referring to the entire iframe.

damargulis: Careful here -- this isn't the "entire" iframe .. this is just the main div in the iframe.  The "entire" iframe is the actual `<iframe></iframe>` element.  All of the code you are writting will be run from inside your iframe.  Therefore if you wanted to access the "entire" iframe, you could call `window.document`.  That would include _everything_, including the actual `<html></html>` element.  

Imo a name like "mainContainer" or "adContainer" might convey that a bit better

juliareichel: okay I will change to adContainer

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    try{
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+        findNearest.classList.add("hidden");})
```
Kaleab470, line 58: minor nit - }) should be on a new line

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    }
+    catch(exception){
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.PAUSE_NOT_HONORED, 
+        message: "Request to pause ad not honored"});
```
Kaleab470, line 62: same here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  }
+
+  /**
+   * Creates the Skip To Content and Return To Ad buttons once the user
+   *   grants permission to access their location and the map appears.
+   * @private 
+  */
+  createButtonsMapState_() {
+    const returnToAdButton = document.createElement("button");
+    returnToAdButton.innerHTML = "Return To Ad";
```
Kaleab470, line 74: use returnToAdButton.textContent here instead of innerHTML

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    returnToAdButton.id = "returnToAd";
+    returnToAdButton.onclick = () => this.playAd_(returnToAdButton); 
+
+    const skipAdButton = document.createElement("button");
+    skipAdButton.innerHTML = "Skip Ad";
```
Kaleab470, line 79: ^same here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    try{
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+        findNearest.classList.add("hidden");})
+    }
+    catch(exception){
```
ryanthompson591, line 60: Instead of using try catch for a promise, there are two ways.

Either make the second parameter a function for when the promise rejects (less good).

Or chain them.
this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(
    () => {
        findNearest.classList.add("hidden");
    }).catch(
    () => {
     /// whatever you're going to do you can either fatal error, or you can log a message.
   });


Here's a good article you could read.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: changed

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: changed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -1,5 +1,7 @@
 
 import BaseSimidCreative from '../base_simid_creative.js';
+import {CreativeMessage} from '../constants.js';
+import {CreativeErrorCode} from '../constants.js';
```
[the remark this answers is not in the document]
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -50,6 +52,54 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    try{
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+        findNearest.classList.add("hidden");})
+    }
+    catch(exception){
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.PAUSE_NOT_HONORED,
```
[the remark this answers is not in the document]
juliareichel: okay, this is a good practice

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: thanks for the link!

src/main/webapp/creatives/maps/simid_map_creative.js
damargulis: Just to add to that -- you want to avoid "Skip To Content", because you don't know for sure that this is the last ad.  Its possible that this ad is part of a pod containing a few ads, a user would be confused if the click "Skip to Content" just to then be shown the second ad in the pod

src/main/webapp/creatives/maps/simid_map_creative.js
damargulis: Nice catch Kaleab!

Try to avoid innerHTML whenever possible.  innerHTML will take whatever you put in as literal HTML, so if you were to for example run the following lines:
```
div.innerHTML = "<div onclick="functionToAttackComputer()">Text</div>";
```
That would be rendered as html, meaning that a click on that div would execute that function.
If you use textContent instead, you would see that as a literal string on the page instead.
In this instance, it probably doesn't matter since you are hardcoding "Return To Ad", but if this came from AdParameters fro example, then it would be very easy for an attacker to inject arbitrary code into your ad, so its good practice to only use innerHTML if you have a very specific need for it

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -84,6 +83,53 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      findNearest.classList.add("hidden");
+    }).catch(() => {
+      this.simidProtocol.reject(eventData, {errorCode: CreativeErrorCode.PAUSE_NOT_HONORED,
```
damargulis, line 90: What is eventData here?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -84,6 +83,53 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    const adContainer = document.getElementById('adContainer');
+    adContainer.appendChild(returnToAdButton);
+    adContainer.appendChild(skipAdButton);
+  }
+
+  /**
+   * Continues to play the ad if user clicks on Return To Ad button.
+   * @private 
+  */
+  playAd_(returnToAdButton) {
```
damargulis, line 122: Just to make sure im clear on this logic -- looks like the skip button only shows up if the user grants their location access, but then if they go back to the ad, the skip button stays visible? Is that all intentional?

juliareichel: Yes that is intentional- when the map appears both buttons return to ad and skip are displayed. If the user clicks return to ad, then I hide that button because it would look weird to have that stay put if they already returned to the ad. However, just because a user returns to the ad, doesn't mean they should now be stripped of the opportunity to skip it, so I continue displaying the skip button for the duration of the ad

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: I am restructuring this to sendMessage as opposed to rejecting so I will not be using eventData

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -84,6 +83,55 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
   */
   grantLocationAccess_() {
-    //ToDo(kristenmason@): implement map
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).then(() => {
+      findNearest.classList.add("hidden");
```
damargulis, line 88: If you're hiding the button either way, might as well hide if before sending the message so you only have to do it once

juliareichel: true, fixed.

juliareichel:
> Lgtm! Can we see a screenshot?

Screenshot 2020-07-08 at 9 54 00 AM - Display 2

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -85,9 +84,58 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
   grantLocationAccess_() {
+    //ToDo(kristenmason@): implement the Google Maps request access functionality
+    findNearest.classList.add("hidden");
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PAUSE).catch(() => {
+      const pauseErrorMessage = {
+        message: "WARNING: Request to pause ad failed",
+      };
+      this.simidProtocol.sendMessage(CreativeMessage.LOG, pauseErrorMessage);
+    });
+    this.createMapState_();
```
damargulis, line 95: Your comments down in the tests say that the map will only load if pause is granted, but it looks like this will create it either way?

juliareichel: Great catch! I should of called createMapState_() only if pauseRequest is resolved. Fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -85,9 +84,59 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    adContainer.appendChild(skipAdButton);
+
     this.loadMap_();
   }
 
+  /**
+   * Continues to play the ad if user clicks on Return To Ad button.
+   * @private 
+  */
+  playAd_(returnToAdButton) {
```
damargulis, line 126: include `@param` notation for this:
```
/**
 * Function description.
 * @param {!Element} returnToAdButton Description of the param.
 * @private
 */
```

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -77,17 +76,71 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
   }
  
   /**
-   * Prompts the users to grant or deny access to their current location.
+   * Requests player to pause, if accepted generate map.
    * @private 
   */
-  grantLocationAccess_() {
+  prepareCreative_() {
+    debugger;
```
damargulis, line 87: remove debuggers

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -2,7 +2,18 @@
...
+const mockSendMessage = jest.fn();
+mockSendMessage.mockResolvedValue(true);
+const mockReject = jest.fn();
+jest.mock('../simid_protocol.js', () => {
+    return jest.fn().mockImplementation(() => {
+    return {sendMessage: mockSendMessage,
+        addListener: () => {},
+        reject: mockReject,
+        resolve: () => {},
+        };
```
damargulis, line 14: fix indentation

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -181,36 +192,55 @@ test('testing that rejection for JSON parsing errors is working', () => {
...
-test('marker is added to map when map loads', () => {
+test('marker is added to map when map loads', async () => {
     const eventData = createInitData();
     testMap.onInit(eventData);
     testMap.onStart(startData);
+
     const findNearestButton = document.getElementById('findNearest');
     findNearestButton.dispatchEvent(new Event('click'));
+
+    const flushPromises = () => new Promise(setImmediate);
```
damargulis, line 227: Can make this a util function in the global scope

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: Deleted

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: fixed

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: done

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: done

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -77,17 +76,69 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
     findNearestButton.innerText = FIND_NEAREST_TEMPLATE_TEXT + buttonLabel;
-    findNearest.onclick = () => this.grantLocationAccess_();
+    findNearest.onclick = () => this.prepareCreative_();
   }
  
   /**
-   * Prompts the users to grant or deny access to their current location.
+   * Requests player to pause, if accepted generate map.
+   * @private 
+  */
```
Kaleab470, line 85: fix indentation here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -77,17 +76,69 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+        const pauseErrorMessage = {
+          message: "WARNING: Request to pause ad failed",
+        };
+        this.simidProtocol.sendMessage(CreativeMessage.LOG, pauseErrorMessage);
+    });
+  }
+
+  /**
+   * Creates the Skip To Content and Return To Ad buttons once the user
+   *   grants permission to access their location and the map appears.
```
Kaleab470, line 101: ^same here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -77,17 +76,69 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  playAd_(returnToAdButton) {
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_PLAY);
+    returnToAdButton.classList.add("hidden");
+    //ToDo(kristenmason@): hide map
+  }
+
+  /**
+   * Returns to video content if user clicks on Skip To Content button.
+   * @private 
+  */
```
Kaleab470, line 137: ^

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -77,17 +76,69 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  playContent_() {
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_SKIP);
+  }
```
Kaleab470, line 140: is this all playContent() does or is there additional stuff that goes in here? If not, everywhere you call playContent() you can basically replace it with this line

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -181,36 +203,52 @@ test('testing that rejection for JSON parsing errors is working', () => {
...
-test('instance of map is instantiated on button click', () => {
+test('instance of map is instantiated if ad paused', async () => {
     const eventData = createInitData();
     testMap.onInit(eventData);
     testMap.onStart(startData);
+
     const findNearestButton = document.getElementById('findNearest');
     findNearestButton.dispatchEvent(new Event('click'));
+
+    await drivePromisesToCompletion();
```
Kaleab470, line 220: was the problem from yesterday b/c the promises were not fully executed before the expect??
Good job figuring that out! That was definitely a hard one to notice

src/main/webapp/creatives/maps/simid_map.html
```diff
@@ -3,10 +3,12 @@
     <link rel="stylesheet" type="text/css" href="simid_map.css">
   </head>
   <body>
-    <div id="button_container">
-      <button id="findNearest"></button>
+    <div id="adContainer">
+      <div id="button_container">
+        <button id="findNearest"></button>
+      </div>
+      <div id="map"></div>
```
kristen-m, line 10: Should the map div go inside of the adContainer here?

kristen-m: Whoops I can't read it is

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -38,8 +60,10 @@ beforeEach(() => {
         }
       };
     SimidProtocol.mockClear();
+    mockReject.mockReset();
     testMap = new SimidMapCreative();
     document.body.innerHTML = `
+    <div id="adContainer"></div>
```
kristen-m, line 66: Is it okay that the button and map aren't nested inside the adContainer div here? Is there any difference in the tests when it's laid out this way?

src/main/webapp/creatives/maps/simid_map_creative.test.js
kristen-m: ^^

src/main/webapp/creatives/maps/simid_map_creative.test.js
```diff
@@ -119,9 +143,9 @@ test('testing that rejection for ad params not found is working', () => {
         type: "SIMID:Player:init",
     };
     testMap.onInit(eventData);
-    
-    const instance = SimidProtocol.mock.instances[0];
-    const rejectMessageObject = instance.reject.mock.calls[0][1];
+
```
kristen-m, line 146: Just a general question but have you looked into using the createInitData function I added in here? It might save you some time from having to repeat the eventData declaration in every test

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: it's neater to break it down into functions, I will include the tracking metrics in here as well

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: that was the issue for fixing kristen's tests, the issue for fixing the other tests was that the request functions had to be mocked by themselves and not as part of the simidProtocol mock. Then using this mocked function, I had to call mockReset so that the mock restarted in every test

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: The tests don't alter the actual html. The only reason I needed to add this div is so that the tests recognized that this adContainer actually exists within the html

src/main/webapp/creatives/maps/simid_map_creative.test.js
juliareichel: thanks for pointing that out! Refactored tests to use that function

PR #31, opened 2020-07-06
Dashboard

### LINKED ISSUE #30: Interaction Dashboard

Kaleab470:
Closes #30 

https://screenshot.googleplex.com/WwsYMRWv6Ri

### DIFF
--- a/src/main/java/com/google/sps/servlets/DashboardServlet.java
+++ b/src/main/java/com/google/sps/servlets/DashboardServlet.java
@@ -11,6 +11,7 @@ import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.sps.servlets.Property;
 
 @WebServlet("/dashboard")
 public class DashboardServlet extends HttpServlet {
@@ -36,7 +37,12 @@ public class DashboardServlet extends HttpServlet {
     int returnToAdCounter = 0;
     
     for (Entity entity : interactions.asIterable()) {
-      System.out.println(entity);
+      findNearestLocationCounter = (boolean) entity.getProperty(Property.FIND_NEAREST_LOCATION) ? ++findNearestLocationCounter : findNearestLocationCounter;
+      grantsLocationCounter = (boolean) entity.getProperty(Property.GRANTS_LOCATION) ? ++grantsLocationCounter : grantsLocationCounter;
+      interactsWithMapCounter = (boolean) entity.getProperty(Property.INTERACTS_WITH_MAP) ? ++interactsWithMapCounter : interactsWithMapCounter;
+      skipToContentCounter = (boolean) entity.getProperty(Property.SKIP_TO_CONTENT) ? ++skipToContentCounter : skipToContentCounter;
+      returnToAdCounter = (boolean) entity.getProperty(Property.RETURN_TO_AD) ? ++returnToAdCounter : returnToAdCounter;
+      
     }
   }
 }
--- a/src/main/java/com/google/sps/servlets/DashboardServlet.java
+++ b/src/main/java/com/google/sps/servlets/DashboardServlet.java
@@ -19,6 +19,19 @@ public class DashboardServlet extends HttpServlet {
   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        
+    Query interactionQuery = new Query(DBUtilities.INTERACTION_TABLE);
+    
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery interactions = datastore.prepare(interactionQuery);
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/sps/servlets/HomeServlet.java
```diff
@@ -0,0 +1,17 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns the dashboard page. */
+@WebServlet("/")
+public class HomeServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html;");
+    response.sendRedirect("creatives/maps/dashboard.html");
```
damargulis, line 15: If sendRedirect works, doesn't that mean that the user can just go directly to <baseurl>/creatives/maps/dashboard.html to see the page?

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.sps.servlets.Property;
+import java.util.HashMap;
+
+@WebServlet("/dashboard")
+public class DashboardServlet extends HttpServlet {
+
+ /**
```
damargulis, line 20: Fix description

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,69 @@
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery interactions = datastore.prepare(interactionQuery);
+
+    int findNearestLocationCounter = 0;
+    int grantsLocationCounter = 0;
+    int interactsWithMapCounter = 0;
+    int skipToContentCounter = 0;
+    int returnToAdCounter = 0;
+    
+    for (Entity entity : interactions.asIterable()) {
```
damargulis, line 40: You can consider looking into stream filter methods to simplify some of this

https://beginnersbook.com/2017/10/java-8-stream-filter/

src/main/java/com/google/sps/servlets/HomeServlet.java
Kaleab470: Yes. I realized that halfway through but I thought the home route taking them there anyway would be easier for them to use. What do you think?

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: Done

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: I used an alternative method here.

Kaleab470:
Updated look: 
https://screenshot.googleplex.com/PGLGyRCSZaL.png

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,75 @@
...
+
+    try {
+      HashMap<String, Double> interactionPercentages = calculatePercentages();
+      
+      String jsonToSend = convertToJson(interactionPercentages);
+
+      response.setContentType("application/json; charset=UTF-8");
+      response.getWriter().println(jsonToSend);
+    
+    } catch (IllegalAccessException exception) {
```
damargulis, line 38: When would this happen?

Kaleab470: The calculatePercentage() method uses the getDeclared function that could possibly throw an exception. Java wouldn't let me compile if i didn't catch that exception so that's why

src/main/java/com/google/sps/servlets/HomeServlet.java
Kaleab470: Would it be better to move the js file associated with it too? or just the html file is fine?

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,75 @@
...
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+    }
+  }
+
+ /**
+  * This function queries for and calculates the percentages of each interaction.
+  * @return A hashmap with the interactions as keys and the percentages as values.
```
kristen-m, line 45: don't forget to include parameters & return here

Kaleab470: hmm, i did include return here and there are no parameters to this method ?

kristen-m: I'm having some trouble reading today haha, I think I saw the HashMap and thought it was parameters for some reason whoops

PR #35, opened 2020-07-09
Additional dashboard

### LINKED ISSUE #34: Additional query features for dashboard

Kaleab470:
https://screenshot.googleplex.com/WEFBPbE4aSG

https://screenshot.googleplex.com/dejvUSrpTjX

https://screenshot.googleplex.com/ZcssJpTSQhk

closes #34


pom.xml
```diff
@@ -68,6 +68,11 @@
       <version>1.23.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
kristen-m, line 71: It looks like some of the commits from your previous PR (#31) are showing up in this one as well, do you need us to approve that PR before you merge this one?

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,107 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    String startTimestamp = RequestUtils.getParameter(request, Property.TIMESTAMP);
+    String endTimestamp;
+
+    if (startTimestamp.isEmpty()) {
+      startTimestamp = DEFAULT_START_TIMESTAMP;
+      endTimestamp = DEFAULT_END_TIMESTAMP;
+    } else {
+      endTimestamp = String.valueOf(Long.valueOf(startTimestamp) + MILLISECONDS_IN_DAY);
```
juliareichel, line 43: is startTime is not empty meaning a user session has begun, why do you set endTime to the start + milisecs in an entire day? Shouldnt endTime be when the user finishes his/her session?

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,107 @@
...
+      response.setContentType("application/json; charset=UTF-8");
+      response.getWriter().println(jsonToSend);
+    
+    } catch (IllegalAccessException exception) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+    }
+  }
+
+ /**
+  * This function queries for and calculates the percentages of each interaction.
```
juliareichel, line 60: add @params

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,107 @@
...
+  * @return A hashmap with the interactions as keys and the percentages as values.
+  */
+  public HashMap<String, Double> calculatePercentages(String startTimestamp, String endTimestamp) throws IllegalAccessException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    final Filter startTimestampFilter =  new FilterPredicate(Property.TIMESTAMP, FilterOperator.GREATER_THAN_OR_EQUAL, startTimestamp);
+    final Filter endTimestampFilter =  new FilterPredicate(Property.TIMESTAMP, FilterOperator.LESS_THAN_OR_EQUAL, endTimestamp);
+    
+    final Query timedQuery = new Query(DBUtilities.INTERACTION_TABLE).setFilter(CompositeFilterOperator.and(startTimestampFilter, endTimestampFilter));
```
juliareichel, line 69: maybe put a small comment about what compositeFilterOperator reutrns (ie: an array of the values) since this is probably not known

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -0,0 +1,107 @@
...
+    PreparedQuery interactions = datastore.prepare(timedQuery);
+
+    Field[] allFields = Property.class.getDeclaredFields();
+  
+    HashMap<String, Double> dataToSend = new HashMap<>();
+
+    // skip the timestamp and correlator properties
+    for (int i = 2; i < allFields.length; ++i){
+      String property = (String) allFields[i].get(new Property()); // gets the value of the field variable
```
juliareichel, line 78: how does new Property() get the value of field variable?

src/main/java/com/google/sps/servlets/InitializeServlet.java
```diff
@@ -26,14 +26,15 @@
...
-    final String correlator = RequestUtils.getCorrelator(request);
+    final String correlator = RequestUtils.getParameter(request, Property.CORRELATOR);
 
     if (correlator.isEmpty()) {
       response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
       return;
     }
 
     Entity interaction = new Entity(DBUtilities.INTERACTION_TABLE);
+    interaction.setProperty(Property.TIMESTAMP, String.valueOf(System.currentTimeMillis()));
```
juliareichel, line 37: We can use this timestamp to not only help with calculating the amount of users who interacted on a given day, but also to see how long each session was. This is a common KPI, which can be extremely helpful for advertisers and also shouldn't be that hard to implement with these new additions.

juliareichel:
> https://screenshot.googleplex.com/dejvUSrpTjX
> When the calendar layout appears, is there a reason the horizontal titles of the graph aren't displayed? This looks a little strange

src/main/webapp/creatives/maps/dashboardScript.js
```diff
@@ -0,0 +1,55 @@
...
+
+    const chart = new google.visualization.ColumnChart(
+        document.getElementById('chart-container'));
+    chart.draw(data, options);
+  });
+}
+
+document.getElementById("form-container").appendChild(createQueryForm());
+
+function createQueryForm(){
```
juliareichel, line 35: perhaps change the name to createQueryByDateForm (to make it clear what you are querying by)

pom.xml
Kaleab470: Oh yes, good catch!

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: for the filter by date query, when the user selects a date, the timestamp that gets sent is the first timestamp of that day. so i'm adding the number of milliseconds in a day to figure out the range of timestamps for that date

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: done

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: it takes in an instance of the object to get the field variables for that class

src/main/webapp/creatives/maps/dashboardScript.js
Kaleab470: done

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -33,14 +33,15 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
-    String startTimestamp = RequestUtils.getParameter(request, Property.TIMESTAMP);
-    String endTimestamp;
+    String startTimestamp = RequestUtils.getParameter(request, "startTime");
+    String endTimestamp = RequestUtils.getParameter(request, "endTime");
 
-    if (startTimestamp.isEmpty()) {
+    // endTimestamp should include the entire second date requested
```
juliareichel: what do you mean by entire "second date"

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -89,12 +90,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       int numUsersInteracted = datastore.prepare(filteredQuery).countEntities();
       int totalInteractions = interactions.countEntities();
 
-      double percentage;
-      if (totalInteractions == 0) {
-        percentage = 0;
-      } else {
-        percentage = numUsersInteracted / (double) totalInteractions;
-      }
+      double percentage = totalInteractions == 0 ? 0 : numUsersInteracted / (double) totalInteractions;
```
juliareichel, line 92: nice use of a ternary operator!

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: when they give a range of dates, i include the latter date as well.

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -113,4 +111,29 @@ private String convertToJson(HashMap<String, Double> data) {
...
+      this.startTimestamp = DEFAULT_START_TIMESTAMP;
+      this.endTimestamp = DEFAULT_END_TIMESTAMP;
+    
+    } else {  
+      // swap if they're in the wrong order
+      if (Long.valueOf(endTimestamp) < Long.valueOf(startTimestamp)) {
+        this.startTimestamp = endTimestamp;
+        this.endTimestamp = startTimestamp;
+      }
+      // endTimestamp should include the entire second date requested
```
juliareichel, line 134: Again what does this comment mean?

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -113,4 +111,29 @@ private String convertToJson(HashMap<String, Double> data) {
...
+  * @param endTimestamp A string containing the requested end time as a UNIX timestamp
+  */
+  private void validateAndSetVariables(String startTimestamp, String endTimestamp){
+
+    if (startTimestamp.isEmpty() || endTimestamp.isEmpty()) {
+      this.startTimestamp = DEFAULT_START_TIMESTAMP;
+      this.endTimestamp = DEFAULT_END_TIMESTAMP;
+    
+    } else {  
+      // swap if they're in the wrong order
```
juliareichel, line 129: why would they be in the wrong order if you are getting these values from the RequestUtils.getParameter

src/main/java/com/google/sps/servlets/DashboardServlet.java
juliareichel: okay so just change the comment to make that clearer "When displaying the range of dates, the latter date is included."

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: only if the user puts in the dates backwards. decided not to address that in newer commits

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
 import java.lang.reflect.Field; 
 import java.util.HashMap;
+import com.google.sps.servlets.RequestUtils;
 
 @WebServlet("/dashboard")
 public class DashboardServlet extends HttpServlet {
 
+  private final long MILLISECONDS_IN_DAY = 86400000;
+  private final String DEFAULT_START_TIMESTAMP = "0";
+  private final String DEFAULT_END_TIMESTAMP = String.valueOf(System.currentTimeMillis());
```
damargulis, line 28: This will make it so that the default end time is determined whenever you first run the app.  I don't think youll be able to use a constant for this

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
+  * Given the start and end timestamps requested, this function validates and sets 
+  * the instance variables.
+  *
+  * @param startTimestamp A string containing the requested start time as a UNIX timestamp
+  * @param endTimestamp A string containing the requested end time as a UNIX timestamp
+  */
+  private void validateAndSetVariables(String requestStartTimestamp, String requestEndTimestamp){
+
+    if (requestStartTimestamp.isEmpty() || requestEndTimestamp.isEmpty()) {
+      this.startTimestamp = DEFAULT_START_TIMESTAMP;
```
damargulis, line 126: There shouldn't be any need to use class variables in something like this.  See if you can refactor it so that you don't need to save anything to the class.  This should be able to process many simultaneous requests. I don't know how using class variables would effect that, but probably not well

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
+    validateAndSetVariables(requestStartTimestamp, requestEndTimestamp);    
+    
     try {
-      HashMap<String, Double> interactionPercentages = calculatePercentages();
+      HashMap<String, Double> dataToSend = calculatePercentages();
       
-      String jsonToSend = convertToJson(interactionPercentages);
+      String jsonToSend = convertToJson(dataToSend);
 
+      response.addHeader("Access-Control-Allow-Origin", "*");
```
damargulis, line 48: Did you confirm that this is actually needed? Ideally, if we can not include this, or at least set it to just the specific domain that needs it, it would be alot better than opening it up to anyone

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
       response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
     }
   }
 
  /**
-  * This function queries for and calculates the percentages of each interaction.
+  * This function queries for and calculates the percentages of each
+  * interaction.
+  *
+  * @param startTimestamp Timestamp for the beginning of the date range that is requested.
```
damargulis, line 63: This function doesn't take in any parameters

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
   */
-  public HashMap<String, Double> calculatePercentages() throws IllegalAccessException {
+  private HashMap<String, Double> calculatePercentages() throws IllegalAccessException {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery interactions = datastore.prepare(new Query(DBUtilities.INTERACTION_TABLE));
+
+    final Filter startTimestampFilter =  new FilterPredicate(Property.TIMESTAMP, FilterOperator.GREATER_THAN_OR_EQUAL, this.startTimestamp);
+    final Filter endTimestampFilter =  new FilterPredicate(Property.TIMESTAMP, FilterOperator.LESS_THAN_OR_EQUAL, this.endTimestamp);
+    
+    // combine the two filters with the 'and' operator
```
damargulis, line 73: This comment is just repeating what the code says, imo its not adding any value

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,125 @@
...
     Field[] allFields = Property.class.getDeclaredFields();
+    Property classInstance = new Property();    
   
-    HashMap<String, Double> interactionPercentages = new HashMap<>();
+    HashMap<String, Double> dataToSend = new HashMap<>();
 
-    for (Field field : allFields) {
-      String property = (String) field.get(new Property()); // gets the value of the field variable
+    // start iterating at 2 to skip timestamp and correlator properties
+    for (int i = 2; i < allFields.length; ++i){
```
damargulis, line 83: This seems bad.  If someone were to add a new parameter, or even just sort the values, this would stop working.  I think it would be better to keep another constant somewhere, thats an array of the fields to search for

src/main/webapp/creatives/maps/UserActivityLogger.js
```diff
@@ -16,7 +16,7 @@ class UserActivityLogger {
    * @param {string} url A URL to send the request to.
    */
   sendRequest_(url) {
-    fetch(`/${url}?correlator="${this.correlator_}"`);
+    fetch(`${URL.BASE_URL}/${url}?correlator=${this.correlator_}`);
```
damargulis, line 19: It reads a bit weird to have `URL.BASE_URL/url`, maybe rename the url param to `path` or similar

src/main/webapp/creatives/maps/utils.js
```diff
@@ -1,13 +1,13 @@
 
 /** A list of server routes that keep track of each interaction when fetched. */
 export const URL = {
```
damargulis, line 3: This seems like it belongs more in a const file then a utils file.
Also, `URL` should be `Urls`.  Its not CAPITAL_CASE since it is an object, so its not deeply immutable.  The properties of it however are deeply immutable and can stay CAPITALIZED
https://google.github.io/styleguide/jsguide.html#naming-constant-names

src/test/java/com/google/sps/servlets/DashboardServletTest.java
```diff
@@ -0,0 +1,103 @@
...
+  @Test
+  public void testCorrectStatusWhenParameterNull() throws IOException {
+
+    // could also be 'endTime' here
+    Mockito.when(request.getParameter("startTime")).thenReturn(null);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
+
+    new DashboardServlet().doGet(request, response);
+
+    // Mockito.when(request.getWriter().println())
```
damargulis, line 59: remove

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: Changed it!

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: Removed it and it still works - i think

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: fixed it

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: removed

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: made a const array of the fields

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: removed class instance variables

src/main/java/com/google/sps/servlets/FindLocationServlet.java
```diff
@@ -32,6 +32,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
     DBUtilities.setToTrue(correlator, Property.FIND_NEAREST_LOCATION);
 
+    response.getWriter().println(DBUtilities.SUCESS_MESSAGE);
```
kristen-m, line 35: Good idea to add this!

src/main/webapp/creatives/maps/UserActivityLogger.js
```diff
@@ -13,33 +14,33 @@ class UserActivityLogger {
 
   /**
    * Makes a GET request to the given URL with the correlator as a parameter.
-   * @param {string} url A URL to send the request to.
+   * @param {string} path A URL to send the request to.
    */
-  sendRequest_(url) {
-    fetch(`/${url}?correlator="${this.correlator_}"`);
+  sendRequest_(path) {
```
kristen-m, line 19: You & Julia have different things written under this function-- do you need to include the mode: 'no-cors' here?

src/main/webapp/creatives/maps/constants.js
```diff
@@ -0,0 +1,11 @@
+
+/** A list of server routes that keep track of each interaction when fetched. */
+export const Urls = {
```
kristen-m, line 3: Could you add the Urls constant to the existing constants.js file that lives in the creatives folder?

src/main/webapp/creatives/maps/dashboardScript.js
```diff
@@ -1,23 +1,24 @@
 
-import {URL} from './utils.js'
-
 google.charts.load('current', {'packages':['corechart']});
 google.charts.setOnLoadCallback(drawChart);
 
+const BASE_URL = "https://step-capstone-team33-2020.appspot.com"
```
kristen-m, line 5: Why assign the project url to BASE_URL here when you created the base url constant earlier? Couldn't you import the constant and use that here as well?

src/main/webapp/creatives/maps/UserActivityLogger.js
Kaleab470: we tested it oiut and it wasn't necessary

src/main/webapp/creatives/maps/constants.js
Kaleab470: oh yes, good call!

src/main/webapp/creatives/maps/dashboardScript.js
Kaleab470: not completely sure how the url parameter thing will work that's why I left it in like that

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,133 @@
...
+      String endTimestamp = String.valueOf(System.currentTimeMillis());
+
+      validatedTimestamps.add(startTimestamp);
+      validatedTimestamps.add(endTimestamp);
+    } else {  
+      String startTimestamp = requestStartTimestamp;
+      String endTimestamp = requestEndTimestamp;
+    
+      // When displaying the range of dates, the latter date is included
+      endTimestamp = String.valueOf(Long.valueOf(endTimestamp) + MILLISECONDS_IN_DAY);
```
damargulis, line 138: But when you make the query, it will actually look at the actual timestamp right, not just the day?  So if they make their request for june 3rd at 4:00, this would then include everything until june 4th at 4:00?

src/test/java/com/google/sps/servlets/DashboardServletTest.java
```diff
@@ -0,0 +1,101 @@
...
+    Mockito.when(request.getParameter("startTime")).thenReturn(null);
+    Mockito.when(response.getWriter()).thenReturn(printWriter);
+
+    new DashboardServlet().doGet(request, response);
+
+    Mockito.verify(response).setStatus(HttpServletResponse.SC_OK);
+  }
+
+  @Test
+  public void testCorrectStatusWhenParameterEmpty() throws IOException {
```
damargulis, line 62: You probably want to verify in these tests that you are handling empty parameters correctly.  So make sure that the query still goes from 0 if you don't include a start, or to the present if you dont include an end

src/test/java/com/google/sps/servlets/DashboardServletTest.java
```diff
@@ -0,0 +1,101 @@
...
+    originalInteraction.setProperty(Property.INTERACTS_WITH_MAP, false);
+    originalInteraction.setProperty(Property.SKIP_TO_CONTENT, false);
+    originalInteraction.setProperty(Property.RETURN_TO_AD, false);
+    originalInteraction.setProperty(Property.TIMESTAMP, "1");
+
+    dataStore.put(originalInteraction);
+
+    new DashboardServlet().doGet(request, response);
+
+    Assert.assertEquals(1, dataStore.prepare(timedQuery).countEntities());
```
damargulis, line 99: Assert that the properties are calculated correctly also

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,133 @@
...
+  *
+  * @param startTimestamp A string containing the requested start time as a UNIX timestamp
+  * @param endTimestamp A string containing the requested end time as a UNIX timestamp
+  * @return An arraylist with two entries - one for each validated start/end time
+  */
+  private ArrayList<String> validateTimestamps(String requestStartTimestamp, String requestEndTimestamp){
+
+    ArrayList<String> validatedTimestamps = new ArrayList<>();
+
+    if (requestStartTimestamp.isEmpty() || requestEndTimestamp.isEmpty()) {
```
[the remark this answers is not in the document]
Kaleab470: I have already taken care of that on the front-end. This is for when everything is queried for so I will change this to an AND

src/main/java/com/google/sps/servlets/DashboardServlet.java
Kaleab470: no, if they choose june 3rd, the timestamp that is sent is for 12:00AM of june 3rd. That's why i add the number of milliseconds in a day to include all of that date

src/test/java/com/google/sps/servlets/DashboardServletTest.java
Kaleab470: that is handled in the front-end if the date picker is empty and it is sent as a normal query with two parameters

src/main/java/com/google/sps/servlets/DashboardServlet.java
```diff
@@ -14,62 +14,133 @@
...
   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
+    String requestStartTimestamp = RequestUtils.getParameter(request, "startTime");
+    String requestEndTimestamp = RequestUtils.getParameter(request, "endTime");
+
+    ArrayList<String> validatedTimestamps = validateTimestamps(requestStartTimestamp, requestEndTimestamp);    
+    String validatedStartTimestamp = validatedTimestamps.get(0);
+    String validatedEndTimestamp = validatedTimestamps.get(1);
```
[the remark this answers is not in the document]
Kaleab470: i thought about doing that but we would need to know both values to see if they should default or not. I am not particularly a fan of how i did it either tbh. maybe its worth doing it all in the doGet method? it's an extra 7 lines - what do you think?

src/test/java/com/google/sps/servlets/DashboardServletTest.java
Kaleab470: Added those!

PR #39, opened 2020-07-13
Added route display functionality

kristen-m:
image

--Displays the shortest route between the user and the closest advertised location
--Gives the user the option to change their transit method through a drop down bar, as pictured (not super pretty right now)
--Allows the user to click on one of the other locations displayed and have a new route be calculated using that location.

### DIFF
--- a/src/main/webapp/creatives/maps/simid_map.css
+++ b/src/main/webapp/creatives/maps/simid_map.css
@@ -43,7 +43,7 @@ html, body {
     width: 80%;
     margin: auto;
     position: relative;
-    top: 50%;
+    top: 215;
     -ms-transform: translateY(-50%);
     transform: translateY(-50%);
 }
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -216,6 +216,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
     };
     const service = new google.maps.places.PlacesService(this.map_);
     service.nearbySearch(request, this.displayResults_.bind(this));
+    this.createTravelChoices_();
   }
 
   /**
@@ -264,7 +265,6 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
    */
   displayDirections_(destination, startingLocation) {
-    this.createTravelChoices_();
     const directionsService = new google.maps.DirectionsService();
     this.directionsRenderer_.setMap(this.map_);
     this.calculateRoute_(directionsService, this.directionsRenderer_, startingLocation, destination);
@@ -285,9 +285,11 @@ export default class SimidMapCreative extends BaseSimidCreative {
     const walkOption = this.createTravelOption_("Walking");
     const driveOption = this.createTravelOption_("Driving");
     const bikeOption = this.createTravelOption_("Bicycling");
+    const transitOption = this.createTravelOption_("Transit");
     travelMethod.add(driveOption);
     travelMethod.add(walkOption);
     travelMethod.add(bikeOption);
+    travelMethod.add(transitOption);
     travelMethod.classList.add("travel-method");
     adContainer.append(travelMethod);
   }
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -259,11 +259,10 @@ export default class SimidMapCreative extends BaseSimidCreative {
    * @private 
    */
   displayDirections_(destination, startingLocation) {
-    const directionsService = new google.maps.DirectionsService();
     this.directionsRenderer_.setMap(this.map_);
-    this.calculateRoute_(directionsService, this.directionsRenderer_, startingLocation, destination);
+    this.calculateRoute_(startingLocation, destination);
     document.getElementById("travel-method").addEventListener("change", () => {
-      this.calculateRoute_(directionsService, this.directionsRenderer_, startingLocation, destination);
+      this.calculateRoute_(startingLocation, destination);
     });
   }
 
@@ -304,17 +303,14 @@ export default class SimidMapCreative extends BaseSimidCreative {
   /**
    * Displays the route between the starting loaction and destination
    * based off of the selected travel mode.
-   * @param {!google.maps.DirectionsRenderer} renderer Object that displays
-   * the directions retrieved by the request.
-   * @param {!google.maps.DirectionsService} service Object that communicates
-   * with the Google Maps API Directions Service.
    * @param {!google.maps.LatLng} start The LatLng coordinates of the start location.
    * @param {!google.maps.LatLng} end The LatLng coordinates of the end location.
    * @private 
    */
-  calculateRoute_(service, renderer, start, end) {
+  calculateRoute_(start, end) {
+    const directionsService = new google.maps.DirectionsService();
     const selectedMode = document.getElementById("travel-method").value;
-    service.route(
+    directionsService.route(
       {
         origin: start,
         destination: end,
@@ -322,7 +318,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
       },
       (response, status) => {
         if (status == "OK") {
-          renderer.setDirections(response);
+          this.directionsRenderer_.setDirections(response);
         } else {
           window.alert("Directions request failed due to " + status);
         }
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -55,15 +55,9 @@ export default class SimidMapCreative extends BaseSimidCreative {
 
     /**
      * A list of the travel modes supported by the Google Maps API.
-     * @private @const {!array}
+     * @private @const {!Array<!string>}
      */
-    this.transportMethods_ = [
-      this.createTravelOption_("Driving"),
-      this.createTravelOption_("Walking"),
-      this.createTravelOption_("Bicycling"),
-      //Maps API designates Transit as bus, train, tram, light rail, and subway.
-      this.createTravelOption_("Transit")
-    ];
+    this.transportMethods_ = [("Driving"),("Walking"),("Bicycling"),("Transit")];
   }
 
   /** @override */
@@ -229,11 +223,11 @@ export default class SimidMapCreative extends BaseSimidCreative {
    */
   displayResults_(results, status) {
     if (status == google.maps.places.PlacesServiceStatus.OK) {
+      this.activeLocation_ = results[0].geometry.location;
       for (let i = 0; i < DEFAULT_LOCATION_NUM_DISPLAYED; i++) {
         this.placeMapMarker_(results[i]);
       }
-      this.activeLocation_ = results[0].geometry.location;
-      this.displayDirections_(this.activeLocation_, this.currentLocation_);
+      this.displayDirections_();
     } else {
       const statusErrorMessage = {
         message: "ERROR: Failed to complete search: "+status,
@@ -261,24 +255,17 @@ export default class SimidMapCreative extends BaseSimidCreative {
     ///Recalculate directions if a different active marker is selected.
     placeMarker.addListener('click', () => {
       this.activeLocation_ = place.geometry.location;
-      this.displayDirections_(this.currentLocation_, this.activeLocation_);
+      this.displayDirections_();
     });
   }
 
   /**
-   * Displays the route between the starting loaction and destination
-   * based off of the selected travel mode.
-   * @param {!google.maps.LatLng} destination The LatLng coordinates of the destination.
-   * @param {!google.maps.LatLng} startingLocation The LatLng coordinates of the start location.
+   * Displays the directions between the user's current location and current active location.
    * @private 
    */
-  displayDirections_(destination, startingLocation) {
+  displayDirections_() {
     this.directionsRenderer_.setMap(this.map_);
-    this.calculateRoute_(startingLocation, destination);
-    //If travel method changes, recalculate directions.
-    document.getElementById("travel_method").addEventListener("change", () => {
-      this.calculateRoute_(startingLocation, destination);
-    });
+    this.calculateRoute_();
   }
 
   /**
@@ -290,10 +277,14 @@ export default class SimidMapCreative extends BaseSimidCreative {
     const travelChoicesContainer = document.getElementById("button_container")
     const travelMethod = document.createElement('select');
     travelMethod.id = "travel_method";
-    for(let i = 0; i < this.transportMethods_.length; i++) {
-      travelMethod.add(this.transportMethods_[i]);
-    }
+    this.transportMethods_.forEach((transportType) =>{
+      const newOption = this.createTravelOption_(transportType);
+      travelMethod.add(newOption);
+    });
     travelMethod.classList.add("travel_method");
+    travelMethod.addEventListener("change", () => {
+      this.calculateRoute_();
+    });
     travelChoicesContainer.append(travelMethod);
   }
 
@@ -311,19 +302,17 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /**
-   * Displays the route between the starting location and destination
-   * based off of the selected travel mode.
-   * @param {!google.maps.LatLng} start The LatLng coordinates of the start location.
-   * @param {!google.maps.LatLng} end The LatLng coordinates of the end location.
+   * Calculates the route between the user's current location and current
+   * active location based off of the selected travel mode.
    * @private 
    */
-  calculateRoute_(start, end) {
+  calculateRoute_() {
     const directionsService = new google.maps.DirectionsService();
     const selectedMode = document.getElementById("travel_method").value;
     directionsService.route(
       {
-        origin: start,
-        destination: end,
+        origin: this.currentLocation_,
+        destination: this.activeLocation_,
         travelMode: [selectedMode]
       },
       (response, status) => {
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -114,6 +114,10 @@ export default class SimidMapCreative extends BaseSimidCreative {
     findNearest.onclick = () => this.prepareCreative_();
   }
 
+  /**
+   * 
+   * @private 
+   */
   prepareCreative_() {
     //ToDo(kristenmason@): implement the Google Maps request access functionality
     findNearest.classList.add("hidden");
@@ -243,8 +247,10 @@ export default class SimidMapCreative extends BaseSimidCreative {
   }
 
   /**
-   * Creates and displays a marker on the map representing a given place.
-   * @param {!Object} place A Place Result object.
+   * Displays the route between the starting loaction and destination
+   * based off of the selected travel mode.
+   * @param {!google.maps.LatLng} destination The LatLng coordinates of the destination.
+   * @param {!google.maps.LatLng} startingLocation The LatLng coordinates of the start location.
    * @private 
   */
   displayDirections_(destination, startingLocation) {
@@ -260,6 +266,11 @@ export default class SimidMapCreative extends BaseSimidCreative {
     });
   }
 
+  /**
+   * Creates a drop down menu where users can choose between
+   * different modes of travel to display directions for.
+   * @private 
+  */
   createTravelChoices_() {
     const adContainer = document.getElementById("button_container")
     const travelMethod = document.createElement('select');
@@ -276,6 +287,18 @@ export default class SimidMapCreative extends BaseSimidCreative {
     adContainer.append(travelMethod);
   }
 
+
+  /**
+   * Displays the route between the starting loaction and destination
+   * based off of the selected travel mode.
+   * @param {!google.maps.DirectionsRenderer} directionsRenderer Object that displays
+   * the directions retrieved by the request.
+   * @param {!google.maps.DirectionsService} directionsService Object that communicates
+   * with the Google Maps API Directions Service.
+   * @param {!google.maps.LatLng} start The LatLng coordinates of the start location.
+   * @param {!google.maps.LatLng} end The LatLng coordinates of the end location.
+   * @private 
+  */
   calculateRoute_(directionsService, directionsRenderer, start, end) {
     const selectedMode = document.getElementById("travel-method").value;
     directionsService.route(
@@ -284,7 +307,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
         destination: end,
         travelMode: [selectedMode]
       },
-      function (response, status) {
+      (response, status) => {
         if (status == "OK") {
           directionsRenderer.setDirections(response);
         } else {
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -169,6 +169,7 @@ export default class SimidMapCreative extends BaseSimidCreative {
  * @private 
  */
   displayMap_(coordinates = new google.maps.LatLng(DEFAULT_MAP_LAT, DEFAULT_MAP_LNG)) {
+    this.currentLocation = coordinates;
     this.map_ = new google.maps.Map(document.getElementById('map'), {
       zoom: DEFAULT_ZOOM,
       center: coordinates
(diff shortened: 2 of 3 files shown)


src/main/webapp/creatives/maps/simid_map.css
```diff
@@ -43,7 +43,7 @@ html, body {
     width: 80%;
     margin: auto;
     position: relative;
-    top: 50%;
+    top: 215;
```
juliareichel, line 46: I just sent out a PR for styling, when that gets approved we should see if this is the proper top for the map

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
-    center: coordinates
-  });
-  new google.maps.Marker({
-    position: coordinates,
-    map: this.map_,
-    title: 'Current Position'
-  });
-  this.findNearby_(this.searchQuery_, coordinates);
-}
+   * Loads a map object that currently defaults to a hardcoded location.
```
juliareichel, line 174: Am I missing something, or does this block of code not do anything different than the above deleted one?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    });
+  }
+
+  /**
+   * Creates a drop down menu where users can choose between
+   * different modes of travel to display directions for.
+   * @private 
+   */
+  createTravelChoices_() {
+    const adContainer = document.getElementById("button_container")
```
juliareichel, line 270: very confusing if you save the const to adContainer when there exists an element with an ID "adContainer". Instead save this const to a different variable name

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    const travelMethod = document.createElement('select');
+    travelMethod.setAttribute("id", "travel-method");
```
juliareichel, line 272: travelMethod.id = "travel-method" is a more succinct way of doing this

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: You aren't missing anything-- it's exactly the same. Github flagged it as different for whatever reason but it's the same functionality

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    const walkOption = this.createTravelOption_("Walking");
+    const driveOption = this.createTravelOption_("Driving");
+    const bikeOption = this.createTravelOption_("Bicycling");
+    const transitOption = this.createTravelOption_("Transit");
```
juliareichel, line 276: is this public transit? Clarify which type of transit this is (trains, boat, subway, bus, etc..)

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    travelMethod.add(driveOption);
+    travelMethod.add(walkOption);
+    travelMethod.add(bikeOption);
+    travelMethod.add(transitOption);
+    travelMethod.classList.add("travel-method");
```
juliareichel, line 281: In our CSS file so far we use CamelCase for the IDs and Classes so lets stay consistent instead of using dashes

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    adContainer.append(travelMethod);
+  }
+
+  /**
+   * Creates an option element representing a mode of travel.
+   * @param {!string} travelMode A string representing the
+   * given mode of travel.
+   * @private 
+   */
+  createTravelOption_(travelMode) {
```
juliareichel, line 291: good use of a helper function!

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
-/**
- * Displays the closest business locations to a user's current location.
- * @param {!Object} results An array of Place Results from the search query.
- * @param {!google.maps.places.PlacesServiceStatus} status The status returned 
- *  by the PlacesService on the completion of its searches.
- * @private 
- */
-displayResults_(results, status) {
+  /**
+   * Displays the closest business locations to a user's current location.
```
juliareichel, line 212: don't specify business because it might be a park etc..

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
-    location: coordinates,
-    name: searchParameter,
-    openNow: true,
-    rankBy: google.maps.places.RankBy.DISTANCE
-  };
-  const service = new google.maps.places.PlacesService(this.map_);
-  service.nearbySearch(request, this.displayResults_.bind(this));
-}
+  /**
+   * Searches for the closest corresponding businesses based off of the given search parameter,
```
juliareichel, line 193: don't specify business because it might be something like a park

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
-  });
-}
+  /**
+   * Creates and displays a marker on the map representing a given place.
+   * @param {!Object} place A Place Result object.
+   * @private 
+   */
+  placeMapMarker_(place) {
+    const placeIcon = {
+      url: this.markerImage_,
```
juliareichel, line 235: In the beginning you set markerImage_ = null but I don't see you change that anywhere in the code. Where do you take the marker parameter (in the adParams) and set that as this url or do you want me to do that?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+   */
+  createTravelOption_(travelMode) {
+    const travelOption = document.createElement("option");
+    travelOption.value = travelMode.toUpperCase();
+    travelOption.text = travelMode;
+    return travelOption;
+  }
+
+  /**
+   * Displays the route between the starting loaction and destination
```
juliareichel, line 299: location* typo

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: I set markerImage_ to the marker param from AdParams on [line 86](https://github.com/googleinterns/step33-2020/pull/39/files/29acb97776d77bf670ae1318bd5ea3f8a1975981#diff-6f605ee0c53a273329bab0887a937b2bR86), I changed these variable names in my last PR so that's probably why it didn't show up in this one!

juliareichel: ah I see! Thanks for pointing me to that change

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Good call

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Fixed

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Fixed

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: The Maps API has transit as an umbrella term for bus, train, tram, light rail, and subway, so unfortunately I can't include each individual option there-- do you think it would be better for the explanation for transit to go as a comment in the code or to rename the variable as publicTransitOption?

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Good call, I changed it!

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: I think I would do both - this way the code specifies in the comments but also users understand when they are choosing their preferred mode

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
-displayResults_(results, status) {
+  /**
+   * Displays the closest business locations to a user's current location.
+   * @param {!Object} results An array of Place Results from the search query.
+   * @param {!google.maps.places.PlacesServiceStatus} status The status returned 
+   *  by the PlacesService on the completion of its searches.
+   * @private 
+   */
+  displayResults_(results, status) {
     if (status == google.maps.places.PlacesServiceStatus.OK) {
```
damargulis, line 219: If the status is not OK, what should happen? 
Can maybe use the LOG feature julia added, but maybe should also resume the video and remove the map?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+      url: this.markerImage_,
+      scaledSize: new google.maps.Size(MARKER_SIZE, MARKER_SIZE)
+    };
+    const placeMarker = new google.maps.Marker({
+      map: this.map_,
+      position: place.geometry.location,
+      icon: placeIcon
+    });
+    placeMarker.addListener('click', () => {
+      this.closestLocation_ = place.geometry.location;
```
damargulis, line 245: If you're doing this, then the `closestLocation_` variable no longer refers to the closest location.  Maybe `activeLocation` or something else to indicate what it is really being used for would be better

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  /**
+   * Creates a drop down menu where users can choose between
+   * different modes of travel to display directions for.
+   * @private 
+   */
+  createTravelChoices_() {
+    const adContainer = document.getElementById("button_container")
+    const travelMethod = document.createElement('select');
+    travelMethod.setAttribute("id", "travel-method");
+    const walkOption = this.createTravelOption_("Walking");
```
damargulis, line 275: You can add all of these travel options to a top-level constant array, then down here, you can just iterate through that array, instead of calling each of these individually

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    travelMethod.add(bikeOption);
+    travelMethod.add(transitOption);
+    travelMethod.classList.add("travel-method");
+    adContainer.append(travelMethod);
+  }
+
+  /**
+   * Creates an option element representing a mode of travel.
+   * @param {!string} travelMode A string representing the
+   * given mode of travel.
```
damargulis, line 290: nit. if the param description goes multiple lines, add a couple spaces to the subsequent lines

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -153,77 +169,156 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+      {
+        origin: start,
+        destination: end,
+        travelMode: [selectedMode]
+      },
+      (response, status) => {
+        if (status == "OK") {
+          this.directionsRenderer_.setDirections(response);
+        } else {
+          window.alert("Directions request failed due to " + status);
```
damargulis, line 320: Don't use window.alert (no one likes popups, even informational ones).
Sending a LOG message could be nice, or possibly displaying something in the ad to the user

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Good call, made that adjustment

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -36,6 +36,22 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+     * The LatLng coordinates representing the user's current location.
+     * @private {?google.maps.LatLng}
+     */
+    this.currentLocation_ = null;
+    /**
+     * The DirectionsRenderer object that displays directions from
+     * the given request.
+     * @private {?google.maps.DirectionsRenderer}
+     */
+    this.directionsRenderer_ = new google.maps.DirectionsRenderer();
```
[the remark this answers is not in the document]
kristen-m: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Addressed in last commit

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: addressed in latest commit

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: done

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+   */
+  findNearby_(searchParameter, coordinates) {
+    const request = {
+      location: coordinates,
+      name: searchParameter,
+      openNow: true,
+      rankBy: google.maps.places.RankBy.DISTANCE
+    };
+    const service = new google.maps.places.PlacesService(this.map_);
+    service.nearbySearch(request, this.displayResults_.bind(this));
```
Kaleab470, line 219: use ES style arrow functions instead of bind here

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+   * based off of the selected travel mode.
+   * @param {!google.maps.LatLng} destination The LatLng coordinates of the destination.
+   * @param {!google.maps.LatLng} startingLocation The LatLng coordinates of the start location.
+   * @private 
+   */
+  displayDirections_(destination, startingLocation) {
+    this.directionsRenderer_.setMap(this.map_);
+    this.calculateRoute_(startingLocation, destination);
+    //If travel method changes, recalculate directions.
+    document.getElementById("travel_method").addEventListener("change", () => {
```
Kaleab470, line 279: addEventListener("change", () => this.calculateRoute_(startingLocation, destination))

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
         this.placeMapMarker_(results[i]);
       }
+      this.activeLocation_ = results[0].geometry.location;
+      this.displayDirections_(this.activeLocation_, this.currentLocation_);
+    } else {
+      const statusErrorMessage = {
+        message: "ERROR: Places Service Status was: "+status,
+      };
+      this.simidProtocol.sendMessage(CreativeMessage.LOG, statusErrorMessage);
+      this.playContent_();
```
Kaleab470, line 242: if the map errors out, wouldn't it make sense to go back to the ad - since they were interested enough to check the map? this is okay as well though

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    const placeIcon = {
+      url: this.markerImage_,
+      scaledSize: new google.maps.Size(MARKER_SIZE, MARKER_SIZE)
+    };
+    const placeMarker = new google.maps.Marker({
+      map: this.map_,
+      position: place.geometry.location,
+      icon: placeIcon
+    });
+    ///Recalculate directions if new marker is clicked.
```
Kaleab470, line 261: what does this mean?

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+  /**
+   * Displays the route between the starting location and destination
+   * based off of the selected travel mode.
+   * @param {!google.maps.LatLng} start The LatLng coordinates of the start location.
+   * @param {!google.maps.LatLng} end The LatLng coordinates of the end location.
+   * @private 
+   */
+  calculateRoute_(start, end) {
+    const directionsService = new google.maps.DirectionsService();
+    const selectedMode = document.getElementById("travel_method").value;
```
Kaleab470, line 322: it looks like you use document.getElementById("travel_method").value a lot. i might be missing something but on default, that element doesn't have anything selected does it? Or does the select tag default to the first transportmethod inserted? if it doesn't this could lead to the value being null.

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +181,162 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+        origin: start,
+        destination: end,
+        travelMode: [selectedMode]
+      },
+      (response, status) => {
+        if (status == "OK") {
+          this.directionsRenderer_.setDirections(response);
+        } else {
+          const directionsErrorMessage = {
+            message: "ERROR: Directions request failed due to " + status,
```
Kaleab470, line 334: does the map status provide enough information as to why it failed? Not completely sure how descriptive it can be for this sentence.

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: I agree with this so return to the ad and then continue to display the skip ad button

juliareichel: Perhaps change message to 'ERROR: Failed to load map: + status "

src/main/webapp/creatives/maps/simid_map_creative.js
juliareichel: 'ERROR: Failed to load directions: + status"

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -36,6 +36,34 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    /**
+     * The DirectionsRenderer object that displays directions from
+     * the given request.
+     * @private @const {!google.maps.DirectionsRenderer}
+     */
+    this.directionsRenderer_ = new google.maps.DirectionsRenderer();
+
+    /**
+     * A list of the travel modes supported by the Google Maps API.
+     * @private @const {!array}
```
damargulis, line 58: `{!Array<!Element>}`

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -36,6 +36,34 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+     */
+    this.transportMethods_ = [
```
damargulis, line 60: The actual strings can be made a top-level const, then you can iterate over them here (or below) to create them buttons

src/main/webapp/creatives/maps/simid_map_creative.js
damargulis, line 279: Does this add a new event listener every time you display directions?
So if you click back and forth between different locations a few times, then pick a new travel method, will calculateRoute_ run a bunch of times in a row?

kristen-m: I'd broken this into 2 lines because keeping it as 1 exceeded the JS column limit from the style guide

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Changed to "ERROR: Failed to complete search: + status"

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: If a different marker is selected on the map, it becomes the new active marker & the directions will be recalculated accordingly. Updated the comment to hopefully be a little more clear, "Recalculate directions if a different active marker is selected."

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: When I add the different options to the select element, the first transportMethod I add has been the selected value by default

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Made this adjustment

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Yes it would get called multiple times-- the goal is to dynamically update the map so that whenever the user makes a change the map reflects that, I wasn't sure if there was another straightforward way to accomplish this

damargulis: Can you add the listener just once when you create the button the first time, then use class variables instead of passing them through?

kristen-m: Great idea-- that's what was causing the bug I found this morning!

src/main/webapp/creatives/maps/simid_map_creative.js
kristen-m: Done

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -36,6 +36,28 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+     * the given request.
+     * @private @const {!google.maps.DirectionsRenderer}
+     */
+    this.directionsRenderer_ = new google.maps.DirectionsRenderer();
+
+    /**
+     * A list of the travel modes supported by the Google Maps API.
+     * @private @const {!Array<!string>}
+     */
+    this.transportMethods_ = [("Driving"),("Walking"),("Bicycling"),("Transit")];
```
damargulis, line 60: This can be top-level, (so not a class variable, but a global one).

Also, no need to put each string in `()`

kristen-m: fixed

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -14,6 +14,7 @@ const DEFAULT_LOCATION_NUM_DISPLAYED = 4;
 const MARKER_SIZE = 25;
 const DEFAULT_MAP_LAT = 37.422004;
 const DEFAULT_MAP_LNG = -122.081402;
+const TRANSPORT_METHODS = ["Driving","Walking","Bicycling","Transit"];
```
ryanthompson591, line 17: nit: add space after comma

src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -154,72 +170,157 @@ export default class SimidMapCreative extends BaseSimidCreative {
...
+    this.directionsRenderer_.setMap(this.map_);
+    this.calculateRoute_();
+  }
+
+  /**
+   * Creates a drop down menu where users can choose between
+   * different modes of travel to display directions for.
+   * @private 
+   */
+  createTravelChoices_() {
```
ryanthompson591, line 271: It just feels like a lot of this stuff should come packaged into the maps API and we shouldn't have to directly specify it.  I mean, what options and choices to we get if we just use the default google maps?

kristen-m: What do you mean by the default google maps? The transport options I have listed here are all of the options offered by the Google Maps API.

PR #45, opened 2020-07-15
Update to send pings when user grants locations

Kaleab470:


src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -86,9 +86,13 @@ export default class SimidMapCreative extends BaseSimidCreative {
     this.userCoordinates_ = adParams[AdParamKeys.COORDINATES];
     const baseUrl = adParams[AdParamKeys.BASE_URL];
 
-    if (baseUrl){
+    if (baseUrl) {
       this.newUserSession_.updateBaseUrl(baseUrl);
     }
+    
+    if (this.userCoordinates_) {
+      this.newUserSession_.userGrantsLocationData();
```
juliareichel, line 94: i dont know if this makes sense because there are default coordinates so how is this a proper way of detecting if user grants location?

kristen-m: The default coordinates should only be created if the displayMap method is called & this.userCoordinates is null, if we perform the check here the only options should be null or the user's real coords!

PR #46, opened 2020-07-15
Update player to display ad properly

kristen-m:
Populate with our default content to make demo more efficient


src/main/webapp/player/simid_player.html
```diff
@@ -120,13 +120,14 @@
       <tr>
         <td>Interactive Creative File</td>
         <td><input type="text" id="creative_url" size = 30 
-           value="../creatives/simid_overlay.html">
+           value="https://step-capstone-team33-2020.appspot.com/creatives/maps/simid_map.html">
```
Kaleab470, line 123: Hopefully, this could save us some repetitive copy-pasting for now. This might have to change back when we put our creative on the github page.

PR #51, opened 2020-07-21
Improved design of creative.

juliareichel:
-Fixed layout of map state
-Fixed button positioning
-Created box for travel time info
-Made info look modular
-Renamed IDs to camelcase to match rest of code

### DIFF
--- a/src/main/webapp/creatives/maps/simid_map.css
+++ b/src/main/webapp/creatives/maps/simid_map.css
@@ -24,7 +24,8 @@ button:hover {
 
 #findNearest{
     position: absolute;
-    bottom: 65px;
+    bottom: 10px;
+    left: 10px;
     padding: 15px;
 }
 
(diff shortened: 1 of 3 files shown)


juliareichel:
image

image

src/main/webapp/creatives/maps/simid_map.css
```diff
@@ -33,23 +34,42 @@ button:hover {
...
+    background: rgba(0,0,0,0.5);
+    border: 1px solid rgba(255,255,255,0.5);
+    color: white;
+    font-size: 18px;
+    margin-left: 11px;
+    top: 60px;
+    position: absolute;
+}
+
+#timeDisplay {
```
Kaleab470, line 47: looks like the box is always positioned to the same spot? If so, what happens when the line connecting the two places passes through that? Might look a little weird.

juliareichel: the line connecting places is centered in the map (it would never be shifted) so this shouldn't be an issue

kristen-m: Kaleab is right-- when I run the code with the design changes the box covers up the route. https://screenshot.googleplex.com/GsHaQXSng0Z, https://screenshot.googleplex.com/CjZDko65jUL. Maybe we could move the travel time box towards the travel method dropdown or I can remove some of the text to make it smaller.

src/main/webapp/creatives/maps/simid_map.css
```diff
@@ -33,23 +34,42 @@ button:hover {
...
+    right: 200px;
     padding: 5px;
-    font-family: Roboto, Arial, sans-serif;
-    font-size: 18px;
+    border: 1px solid black;
+    background-color: rgba(0,0,0,0.6);
+    color: white;
+    max-width: 110px;
+    text-align: center;
+    transition: opacity .5s cubic-bezier(0.0,0.0,0.2,1);
```
damargulis, line 57: had to look up cubic-bezier ---very cool!

src/main/webapp/creatives/maps/simid_map.css
damargulis: Hmm is there a way to use the api to know where the route is being placed? Can you then use that to move the box around dynamically?

juliareichel: tried to find a way but can't seem to find one using the api

damargulis: Hmm, it is somewhat transparant and they can move the map so probably not a huge deal, but up to you all where you think the best place to put it is so that its as unobtrusive as possible

juliareichel: Right I made it transparent for that reason

juliareichel: Kristen is going to shorten the amount of words which should help so this design seems okay

PR #54, opened 2020-07-22
Update player with map creative button 

kristen-m:
--The allow location access button is only displayed if the map creative is selected

### DIFF
--- a/src/main/webapp/player/simid_player.html
+++ b/src/main/webapp/player/simid_player.html
@@ -47,6 +47,9 @@
       document.getElementById('creative_url').value = '../creatives/' + value;
       // populate ad parameters with some hard coded values
       document.getElementById('geolocation').style.visibility = 'hidden';
+      if (value == 'maps/simid_map.html') {
+        document.getElementById('geolocation').style.visibility = 'visible';
+      }
       let adParams = '';
       switch (value) {
         case 'simid_overlay.html':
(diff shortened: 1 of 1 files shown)


src/main/webapp/player/simid_player.html
```diff
@@ -61,6 +61,9 @@
           ];
           adParams = JSON.stringify(adParamsObj);
           break;
+        case 'maps/simid_map.html':
```
damargulis, line 64: If you select simid_map, but then select something else again, doesn't this button stay visible?

kristen-m: whoops, yep. Fixed that

src/main/webapp/player/simid_player.html
```diff
@@ -61,6 +62,9 @@
...
+          document.getElementById('geolocation').style.visibility = 'visible';
+          break;
```
ryanthompson591, line 67: other code changes adParams here. Do you need to do that. If not consider emptying that field.

src/main/webapp/player/simid_player.html
```diff
@@ -46,6 +46,10 @@
     window.switchCreative = (value) => {
       document.getElementById('creative_url').value = '../creatives/' + value;
       // populate ad parameters with some hard coded values
+      document.getElementById('geolocation').style.visibility = 'hidden';
+      if (value == 'maps/simid_map.html') {
```
damargulis, line 50: Keep this as part of the switch, no need to have an extra if statement if you're gonna use a switch anyway.

kristen-m: Fixed

src/main/webapp/player/simid_player.html
kristen-m: fixed

PR #55, opened 2020-07-24
Updated media for park ad

kristen-m:


src/main/webapp/player/simid_player.html
```diff
@@ -138,8 +138,7 @@
       <tr>
         <td>Media File</td>
         <td><input type="text" id="video_url" size = 30 
-           value="https://storage.googleapis.com/gvabox/media/red30s.mp4 
-           ">
+           value="http://storage.googleapis.com/gvabox/damargulis/step/np_ad.mp4">
```
ryanthompson591, line 141: This is fine here, if we eventually want to merge this into the IAB stuff, we should use local mp4 files where possible.

PR #56, opened 2020-07-27
fix small bug

Kaleab470:
Fix issue of time travel and timedisplay not disappearing after returning to ad

### DIFF
--- a/src/main/webapp/creatives/maps/simid_map_creative.js
+++ b/src/main/webapp/creatives/maps/simid_map_creative.js
@@ -11,6 +11,8 @@ const AdParamKeys = {
   BASE_URL: 'baseUrl',
 };
 
+const TRAVEL_METHOD = "travelMethod";
+const TIME_DISPLAY = "timeDisplay";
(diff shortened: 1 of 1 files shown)


src/main/webapp/creatives/maps/simid_map_creative.js
```diff
@@ -207,8 +207,8 @@ export default class SimidMapCreative extends BaseSimidCreative {
     returnToAdButton.classList.add("hidden");
     const mapDiv = document.getElementById("map");
     mapDiv.classList.add("hidden");
-    const travelDisplay = document.getElementById("travel_method");
-    const timeDisplay = document.getElementById("time_display");
+    const travelDisplay = document.getElementById("travelMethod");
```
damargulis, line 210: These are being created dynamically right? So the id is also set in javascript, not in html?  If thats true, you can replace these with a constant and avoid this kind of error.

Kaleab470: Done!

